PR #44, opened 2026-07-16
Model-agnostic local co-evolution: CPU prompt-level self-improvement + Ollama discovery

SHA888:
## Summary

Makes EVOSEAL's co-evolution model-agnostic and runnable on a CPU-only host. Instead of the hardcoded, uninstalled `devstral:latest`, models are **discovered from Ollama by family**, and a new **prompt-level** self-improvement loop replaces GPU-only weight fine-tuning as the default path.

## What's included

- **`fix(config)`** — Repairs a pre-existing pydantic v2 `model_validator` bug that made `evoseal.providers` unimportable in any faithful install (it was written as an instance method under `mode="before"`, so pydantic passed a `ValidationInfo` where a dict was expected). Adds `evoseal/providers/local_models.py`: per-role Ollama model discovery matched by family (coder/reviewer), with `EVOSEAL_CODER_MODEL` / `EVOSEAL_REVIEWER_MODEL` overrides and an offline fallback.
- **`refactor(fine-tuning)`** — Renames `DevstralFineTuner` → `ModelFineTuner` (deprecated alias kept), resolves the coder model from discovery when none is given, and chooses the HF base model from a family map instead of assuming Mistral. De-Devstral's docstrings/health-checks across `evolution/`, `services/`, and dev scripts.
- **`feat(prompt-evolution)`** — New `evoseal/prompt_evolution/`: a coder model writes code, a reviewer model critiques + scores it, and the critique is distilled into a **regression-gated** edit of the coder's system prompt (re-run the task; adopt only if the score improves, else roll back). Guardrails prevent a degenerate rewrite or a plain critique from replacing a working prompt. Includes 23 unit tests.
- **`docs`** — README + Phase 3 / deployment / systemd / API docs reframed around two paths (prompt-level CPU default vs weight-level GPU-only), plus a new `docs/architecture/local_coevolution.md` with a sequence diagram.
- **`test(budget)`** — Fixes a pre-existing token-accumulation test that under-recorded a cycle (asserted 8200, actual 8100); implementation was correct.

## Verification

- 23 new `prompt_evolution` unit tests pass; `test_budget_exhaustion.py` (10 tests) passes.
- Discovery verified live: coder → `deepseek-coder-v2:16b`, reviewer → `qwen2.5-coder:7b`.
- Live smoke on real Ollama exercised the generate → review → evolve → revalidate → **rollback** path end-to-end.
- `ruff` clean across `evoseal/`; new modules type-clean under `mypy`.

## Notes

- The weight-fine-tuning path (`evoseal/fine_tuning/`) is unchanged in behavior and remains GPU-only; it's now model-agnostic and documented as such.
- Per repo convention, commits use Conventional Commits and omit `Co-Authored-By` trailers.

## Summary by CodeRabbit

* **New Features**
  * Added CPU-friendly prompt-level co-evolution using local coder/reviewer models, with guarded prompt updates and rollback when improvements don’t hold.
  * Introduced role-based Ollama model discovery plus new environment variables to select coder/reviewer models, alongside a reviewer-specific provider.
  * Generalized GPU fine-tuning into a model-agnostic `ModelFineTuner` (keeping a deprecated Devstral alias).

* **Documentation**
  * Updated Phase 3, architecture, deployment, and troubleshooting to match role-based local models and the CPU vs GPU workflow split.

* **Bug Fixes**
  * Improved resilience when loading pipeline state, and hardened enum/health-check related behavior.

* **Tests**
  * Expanded prompt-evolution unit tests and updated integration/CI coverage and skipping behavior.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -7,3 +7,8 @@
   # Ollama (only needed if not using localhost:11434 default)
   OLLAMA_API_BASE=http://localhost:11434
   EVOSEAL_DEFAULT_PROVIDER=ollama
+  # Local co-evolution models (CPU path). Coder writes, reviewer critiques.
+  # Pull with: ollama pull deepseek-coder-v2:16b-lite-instruct-q8_0
+  #            ollama pull qwen2.5-coder:7b-instruct-q6_K
+  EVOSEAL_CODER_MODEL=deepseek-coder-v2:16b-lite-instruct-q8_0
+  EVOSEAL_REVIEWER_MODEL=qwen2.5-coder:7b-instruct-q6_K
--- a/.env.example
+++ b/.env.example
@@ -1,14 +1,16 @@
 # Anthropic
-  ANTHROPIC_API_KEY=sk-ant-
+ANTHROPIC_API_KEY=sk-ant-
 
-  # OpenAI
-  OPENAI_API_KEY=sk-proj-
+# OpenAI
+OPENAI_API_KEY=sk-proj-
 
-  # Ollama (only needed if not using localhost:11434 default)
-  OLLAMA_API_BASE=http://localhost:11434
-  EVOSEAL_DEFAULT_PROVIDER=ollama
-  # Local co-evolution models (CPU path). Coder writes, reviewer critiques.
-  # Pull with: ollama pull deepseek-coder-v2:16b-lite-instruct-q8_0
-  #            ollama pull qwen2.5-coder:7b-instruct-q6_K
-  EVOSEAL_CODER_MODEL=deepseek-coder-v2:16b-lite-instruct-q8_0
-  EVOSEAL_REVIEWER_MODEL=qwen2.5-coder:7b-instruct-q6_K
+# Ollama (only needed if not using localhost:11434 default)
+OLLAMA_API_BASE=http://localhost:11434
+EVOSEAL_DEFAULT_PROVIDER=ollama
+
+# Local co-evolution models (CPU path). Coder writes, reviewer critiques.
+# Models are auto-discovered from Ollama by family; set these only to pin a tag.
+# Pull with: ollama pull deepseek-coder-v2:16b-lite-instruct-q8_0
+#            ollama pull qwen2.5-coder:7b-instruct-q6_K
+EVOSEAL_CODER_MODEL=deepseek-coder-v2:16b-lite-instruct-q8_0
+EVOSEAL_REVIEWER_MODEL=qwen2.5-coder:7b-instruct-q6_K
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,7 +248,7 @@ jobs:
       uses: astral-sh/setup-uv@v5
 
     - name: Install documentation dependencies
-      run: uv pip install --system -r requirements/docs.txt
+      run: uv pip install --system -e ".[docs]"
 
     - name: Build documentation
       run: |
@@ -324,9 +324,13 @@ jobs:
       run: uv pip install --system safety
 
     - name: Run Bandit security linter
-      uses: PyCQA/bandit@main
-      with:
-        args: -r evoseal/ -n 1 --skip B101,B104
+      # PyCQA/bandit is the bandit source repo, not a GitHub Action (it has no
+      # action.yml), so `uses:` always failed. Run the tool directly instead.
+      # Gate on medium+ severity; the low findings here are subprocess/random
+      # usage and a "password" false positive on a 0.7 threshold constant.
+      run: |
+        uv pip install --system bandit
+        bandit -r evoseal/ -n 1 --skip B101,B104 --severity-level medium
 
     - name: Run Safety check
       run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -447,7 +447,7 @@ jobs:
 
       - name: Run Trivy vulnerability scan (image)
         if: github.event_name != 'pull_request'
-        uses: aquasecurity/trivy-action@0.34.0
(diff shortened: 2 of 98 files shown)


SHA888:
## 🤖 Automated review — local reviewer agent (Qwen2.5-Coder 7B)

Reviewed by our OpenClaw **reviewer agent** (`qwen2.5-coder`, CPU/Ollama, 16k ctx) over the four core new files. Its raw findings are below, each annotated with a maintainer triage (the model is a 7B local reviewer, so several are false positives — kept here for transparency).

### `evoseal/prompt_evolution/prompt_evolver.py`
- _"`_extract` doesn't handle multiple blocks where none contain the protected markers."_ → **Already handled**: `_extract` iterates blocks via `_looks_like_prompt` (requires markers), else falls through to `None`.
- _"`_validate` doesn't reject candidates that add new protected markers."_ → **By design**: markers must be *retained*, not prevented from being added.
- _"`_summarize_reason` doesn't handle no-issues-but-low-score."_ → **Already handled**: `issues[0] if issues else (summary or "low score")`.

### `evoseal/prompt_evolution/coevolution_manager.py`
- _"`default_manager` doesn't set `accept_threshold`/`min_score_gain`/`coder_seed_prompt`."_ → **Intentional**: the factory relies on the constructor defaults.
- _"`_parse_score` doesn't handle `85/100`."_ → **Already handled**: the second regex matches `/100` and normalizes `value/maximum`.
- _"`_extract_code` doesn't handle malformed code blocks."_ → **No change**: suggested fix is identical to current code.

### `evoseal/prompt_evolution/prompt_store.py`
- _"`register` doesn't validate that a supplied `parent_id` exists."_ → **Valid (minor hardening)**. In practice `parent_id` defaults to the active version so it's always real, but an explicit unknown `parent_id` is silently accepted. Worth a guard.
- _"`rollback` doesn't handle no active prompt."_ → **Already handled**: `if active is None or not active.parent_id: return active`.
- _"`list_versions` can return `None`."_ → **Already handled**: comprehension filters `if v is not None`.

### `evoseal/providers/local_models.py`
- _"`urlopen` without SSL verification is insecure (S310)."_ → **N/A**: it's an `http://localhost` call to Ollama; TLS doesn't apply. The `# noqa` is appropriate.
- _"`select_model` doesn't early-return on empty `available`."_ → **Cosmetic**: current flow already returns `None` correctly.
- _"`resolve_role_models` doesn't guard empty `overrides`/`roles`."_ → **Already handled**: `overrides or {}`; empty roles yields an empty dict.

### Maintainer summary
**1 actionable item** out of 12: harden `PromptStore.register` to validate an explicitly-passed `parent_id`. The remaining findings are false positives or by-design (typical for a 7B local reviewer). No correctness or security defects confirmed.

SHA888:
## Review findings addressed (`de58c61`)

All findings from the self-review and CodeRabbit are fixed. CI is green; suite is **596 passed / 9 skipped** (+32 tests).

### Correctness
- **`EVOSEAL_CODER_MODEL` / `EVOSEAL_REVIEWER_MODEL` did nothing.** They were documented in five places, but `resolve_model()`'s `override` parameter was never passed by any caller — the documented flag was silently dead. Now resolved as *explicit arg → env → family preference*, with tests. Verified live.
- **Float comparison** (CodeRabbit): a candidate exactly meeting `min_score_gain` was rolled back. Fixed with an epsilon.

  ⚠️ One correction: the specific example in the review (`0.3 + 0.05 == 0.35000000000000003`) is **not** true in Python — that sum is exactly `0.35`, so a test built on it would be vacuous. The finding is still valid; the reachable cases come via an independent `value/maximum` division (e.g. `0.05 + 0.1 == 0.15000000000000002`, reachable from a `SCORE: 15/100` review — 781 such pairs). The regression test uses a real case and **fails without the fix**.

### Stability / performance
- No blocking HTTP at import: `SEALConfig`'s `default_factory` called `resolve_model()` twice, so `import evoseal.config` hit Ollama. Now passes `role` and resolves lazily in `OllamaProvider` (CodeRabbit's suggested shape).
- `list_installed_models()` is cached (`clear_model_cache()` to refresh) — it was re-querying per provider/tuner/validator construction and per `create_evolution_result()`, i.e. blocking I/O in async paths.
- A provider error in generate/review killed `run_cycle()` while `evolve()` was already defensive; it now returns a failed cycle (`CoevolutionCycleResult.failed`).

### Safety
- Dropped the "any installed model" fallback — with only an embedding model pulled it would have been selected to write code. Falls back to the known-good tag instead.
- `PromptStore` now writes `registry.json` atomically (temp + `os.replace`) and uses a monotonic sequence instead of `len(versions)` (which collides if a version is pruned). This is the file rollback depends on.

### Tooling
- `.env.example` keys were indented, breaking `source .env`.
- `release.yml` → repointed at `scripts/lib/version/version.py`.
  Note: `bump_version.py` already exited 1 without `--force` **before this PR**, so the release bump step was already broken — not a regression. Chasing it surfaced that the *replacement* was broken too: its `ROOT_DIR` resolved to `scripts/lib/`, so `bump` couldn't find `pyproject.toml`, and it needs `tomli-w`, which the workflow never installed. Both fixed and verified with `--dry-run` (`Would update from 0.3.7 to 0.3.8`).
  Caveat: `release.yml` only runs on `workflow_dispatch`/tags, so this is dry-run-verified, not exercised by a real release.
- `trivy-action@0.34.0` → `@v0.34.0` (the tag is published with a leading `v`).

### Known, not addressed (deliberately)
`tests/integration/seal/test_knowledge_base_concurrent.py::test_concurrent_access` flakes under `-n auto` (passes alone; it asserts a strict wall-clock timeout around a `ThreadPoolExecutor`). Pre-existing and unrelated — filed separately rather than growing this PR further.

PR #58, opened 2026-07-19
feat: wire evoseal start evolution CLI + fix CLI dispatch bug

SHA888:
## Summary

- Adds `evoseal start evolution`, which constructs and runs the Phase 3 `ContinuousEvolutionService` (lazy import to keep the CLI GPU-optional). Prints an explicit research-stage notice since the bidirectional co-evolution loop it launches is not yet closed. Ticks the "Wire the loop into the CLI" item in TODO.md's co-evolution backlog.
- **Fixes a real, pre-existing bug**: `evoseal/cli/__init__.py` had dead duplicate-registration code that re-registered the app's top-level callback with a bare `lambda version: None`, clobbering the working `--version` option with a required positional argument. This broke dispatch for *every* `evoseal <subcommand> <args>` invocation — only bare `--help` happened to work. The one existing CLI-dispatch test was a false positive (it asserted `"doctor"` appears in `evoseal doctor --help` output, which passed because the broken app printed its own top-level help, listing "doctor" in the commands table, instead of doctor's actual help).
- **Fixes a test-isolation bug** found while diagnosing the above: fixtures across the suite that build isolated git repos in temp dirs (via subprocess or GitPython) don't pass an explicit `env=`, so if a parent process sets `GIT_DIR`/`GIT_WORK_TREE`/`GIT_INDEX_FILE` (pre-commit does this while stashing unstaged changes during a commit), those fixtures get silently redirected onto the real repo instead of the isolated temp dir. Extended the existing session-scoped `_git_identity` autouse fixture in `tests/conftest.py` to strip these vars for the whole test session.
- **Fixes the pre-push `trufflehog` hook**: pinned to v3.66.0, whose CLI takes `--[no-]entropy` as a boolean switch; the configured `--entropy=False` was invalid syntax, so the hook has never actually run successfully.
- Repo-wide hygiene fixes surfaced by `pre-commit run --all-files`: one metrics JSON file had invalid JSON (unquoted string literal in an array), two benchmark JSON files were missing a trailing newline.

## Test plan

- [x] `uv run pytest tests/unit/test_start_evolution_command.py -q` — 3/3 pass
- [x] Full suite via `./scripts/evoseal test` — 467/467 pass
- [x] Reproduced the CLI dispatch bug (`evoseal doctor`, `evoseal start api` etc. all failed with "No such command") and confirmed the fix resolves it for every subcommand
- [x] Reproduced the GIT_DIR leakage bug by setting `GIT_DIR`/`GIT_WORK_TREE` to this repo before running the suite; confirmed all 467 tests pass under that condition after the fix
- [x] `pre-commit run --all-files` — all hooks green, including a real `trufflehog` scan (previously silently broken)
- [x] `ruff check` / `black --check` / `isort --check-only` clean on all changed files

## Summary by CodeRabbit

* **New Features**
  * Added a new `evolution` CLI command to start the continuous evolution service.
  * Added `--data-dir`, `--evolution-interval`, and `--training-check-interval` options.
  * Includes graceful stop behavior on interruption.
* **Bug Fixes**
  * Fixed CLI dispatch/registration behavior for the help/version flow.
  * Corrected formatting in metrics/benchmark JSON and updated secret-scan baseline records.
* **Documentation**
  * Updated TODO progress tracking for the evolution workflow.
* **Tests**
  * Added unit tests covering command output, option forwarding, service startup, and interruption handling.


SHA888:
Addressed CodeRabbit review feedback (2 inline comments):

**1. TruffleHog  flag () — SKIPPED**
CodeRabbit claimed `--no-entropy` would fail with an unknown-option error. Verified locally against trufflehog v3.66.0 (the exact pinned version): `trufflehog git --help` shows `--[no-]entropy` as a valid backwards-compat no-op flag, and the command runs successfully (exit 0). No change needed.

**2. Missing pytest unit marker (`tests/unit/test_start_evolution_command.py:21`) — APPLIED**
Added `pytestmark = pytest.mark.unit` using the repo's convention (module-level variable, not decorator — matching `test_openevolve_adapter_remote.py`, `test_dgm_adapter_remote.py`, etc.). Verified: ruff clean, 467/467 tests pass.

SHA888:
## Review feedback — addressed

Verified each point against the actual repo state. Here's the verdict:

### Points 1, 2, 4 — already handled / incorrect (no change needed)

**`evoseal/cli/__init__.py` registration removal** — `main.py` independently owns all subcommand registrations. Verified:
- Console script entry point is `evoseal.cli.main:app` (pyproject.toml), NOT `evoseal.cli:app`
- `evoseal/cli/main.py` already had all registrations on `origin/main` before this PR (lines 55–66), plus `doctor`, `pipeline`, `estimate_cost` that the old `__init__.py` didn't have
- The old block was genuinely broken: `app.callback(invoke_without_command=True)(lambda version: None)` clobbered the working `version_callback` with a bare lambda

**`main.py::main()` no-args-is-help** — `no_args_is_help=True` is set on both `typer.Typer(...)` (line 47) and `@app.callback(...)` (line 79), both on `origin/main` before this PR. The removed `if len(sys.argv) == 1` branch was redundant. `import sys` is not dead — used at lines 32–33 (`sys.path.insert`).

**`TODO.md` coverage** — The other subcommands already have pre-existing test coverage: `tests/unit/test_doctor_command.py`, `tests/unit/test_estimate_cost_command.py`, `tests/integration/openevolve/test_openevolve_integration.py` (covers `init`). "Wire the loop into the CLI" was about adding `start evolution`, not re-testing pre-existing commands.

### Point 3 — applied ✅

**`start_evolution` error handling** — Added `logger.exception("Evolution service failed")` before the user-facing `typer.echo`. Full tracebacks now go to the log; the clean CLI message still goes to the user. Commit `12045ae`.

### Verification

- `ruff format --check` / `ruff check` — clean
- `pytest tests/unit/test_start_evolution_command.py` — 4/4 pass
- Full suite — 468/468 pass

PR #67, opened 2026-07-21
feat: wire daemon _run_evolution_cycle to real EvolutionPipeline

SHA888:
## What

Wire `ContinuousEvolutionService._run_evolution_cycle` to actually invoke `EvolutionPipeline.run_evolution_cycle()` instead of only reading `get_statistics()` and simulating.

## Changes

- **continuous_evolution_service.py**: Added `EvolutionPipeline` import, optional `pipeline` constructor param, and `_get_pipeline()` lazy initializer. `_run_evolution_cycle` now calls `pipeline.run_evolution_cycle(iterations=1)` and logs per-iteration success/failure. Exceptions are caught so a pipeline failure does not crash the service.
- **test_continuous_evolution_service.py**: 4 new tests covering pipeline invocation, failure handling, exception safety, and lazy initialization.
- **TODO.md**: Flipped checkbox item 2 from `[ ]` to `[x]` with date note; updated tracking table.

## Why

This is the first concrete step in closing the bidirectional co-evolution loop. The daemon was simulating evolution instead of running it. Steps 3-5 (model validation, deployment, registry consultation) depend on this wiring being in place.

## Verification

- `ruff format --check .` — pass
- `ruff check evoseal/ tests/` — pass
- `pytest tests/unit/services/test_continuous_evolution_service.py` — 6/6 passed
- Full test suite pre-commit: 472 passed

Addresses TODO.md item: **Wire daemon to real EvolutionPipeline**

## Summary by CodeRabbit

* **New Features**
  * Continuous evolution cycles now run through the real evolution pipeline with configurable iteration counts.
  * Supports either injecting a pipeline instance or using automatic lazy initialization.
* **Bug Fixes**
  * Improved handling and reporting of pipeline and configuration errors without interrupting service operation.
  * Training-readiness now triggers based on the number of successfully completed evolution records.
  * Pipeline results are persisted only when they meet fitness/required code requirements.
* **Documentation**
  * Updated the EVOSEAL improvement plan and tracking totals.
* **Tests**
  * Expanded unit coverage for pipeline looping, persistence rules, error paths, and readiness gating.
* **Chores**
  * Added `psutil` for production runtime needs.


SHA888:
Addressed review feedback from CodeRabbit.

**What changed** (`tests/unit/services/test_continuous_evolution_service.py`):
- Converted the four `_run_evolution_cycle` tests from sync `asyncio.run()` to `async def` + `await`, matching pytest-asyncio strict mode requirements (`asyncio_mode = strict` in pytest.ini)
- Added `@pytest.mark.unit` to all four tests
- Added `@pytest.mark.asyncio` to the three async-dependent tests (invokes pipeline, handles failure, handles exception)
- `test_run_evolution_cycle_lazy_pipeline` correctly gets only `@pytest.mark.unit` — it calls the synchronous `_get_pipeline()` method and does not need async

**Note:** The pre-existing `_trigger_training_cycle` tests (lines 1–69) use the same `asyncio.run()` pattern and lack markers. Left untouched since they weren't part of this PR's diff.

**Verification:**
- `ruff format --check .` — pass
- `ruff check evoseal/ tests/` — pass
- `pytest tests/unit/services/test_continuous_evolution_service.py` — 6/6 passed
- Full suite: 472 passed

SHA888:
## Review feedback addressed

### ✅ Fixed: per-item try/except in `_run_evolution_cycle`

Pushed `4f2a0a0`. The per-result body inside the `for result in results:` loop is now wrapped in its own `try/except Exception` so one malformed item (e.g. `result.get("metrics")` returning `None`, causing `TypeError` on `"fitness" not in metrics`) logs a warning and continues the batch instead of aborting the entire cycle. Added `test_run_evolution_cycle_malformed_item_does_not_abort_batch` to cover the scenario — the good result after the bad one still gets persisted, and cycle-level stats stay consistent.

### ✅ Verified: pipeline return contract

Checked `EvolutionPipeline._run_single_iteration()` (lines 717–897). The real return shape is:

- **Success:** `{iteration, success: True, is_improvement, metrics, should_continue, resilience_events, resilience_status}`
- **Failure:** `{iteration, success: False, metrics: {}, should_continue, error, error_type, recovery_attempted, resilience_status}`

Confirmed: the pipeline never sets `original_code` or `improved_code`. The code already guards against this with explicit comments ("EvolutionPipeline doesn't return a code diff yet") and `continue` statements — so no real result will be silently persisted as a codeless placeholder. The guards are correct and intentional. The `data_collector` will accumulate nothing until the pipeline starts returning code diffs, which is the expected behavior given the current pipeline state. No code change needed.

### ℹ️ Acknowledged: `psutil` scope creep

Agreed — the `psutil` move from `test` extra to core `dependencies` is a valid fix (unconditional import in `evoseal/core/testrunner.py`) but unrelated to this PR's scope. Will keep dependency-classification fixes in separate PRs going forward.

SHA888:
Addressed all three review points. Here's the verification against actual code:

**1. `or` → `and` in codeless-result guard** ✅ Fixed

Confirmed the bug: `not (original_code or improved_code)` only skips when *both* are empty. A result with `original_code="x=1"` and `improved_code=""` would pass through and be persisted — no actual diff to fine-tune on. Changed to `not (original_code and improved_code)` so the guard requires both sides.

**2. Per-item logging outside the `try` block** ✅ Fixed

Confirmed the bug: `result.get("success")` runs before the `try`, so a non-dict entry (None, str, int) raises `AttributeError` that propagates to the outer `except`, aborting the whole batch. Moved logging inside the `try` and added an `isinstance(result, dict)` check with a skip+log for non-dict entries.

**3. Silent empty `seal_config` degradation** ✅ Acknowledged (warning improved)

Confirmed that `EvolutionConfig` defaults `seal_config` to `{}`, so the pipeline won't fail at construction. But downstream code may expect specific keys. Improved the warning message to hint at this: "If the pipeline requires seal_config keys at runtime, it will fail deep inside the run rather than here."

**Tests added:**
- `test_run_evolution_cycle_skips_half_code_results` — verifies the `and` guard rejects results with only original_code
- `test_run_evolution_cycle_non_dict_item_does_not_abort_batch` — verifies non-dict entries (None, str, int) are skipped without crashing the batch

All 18 tests pass. `ruff format` and `ruff check` clean.

SHA888:
## Review feedback addressed (commit 0ab0d01)

### 1. ✅ Narrowed `except TypeError` to `_get_pipeline()` only

The old `except TypeError` wrapped the entire try body — pipeline call, batch loop, stats update. A real bug inside `pipeline.run_evolution_cycle()` or the result loop that raised TypeError would be misreported as `Configuration error... inject a pre-built pipeline`, masking the actual fault.

**Fix:** Split into two try/except blocks. The outer one catches TypeError only from `self._get_pipeline()` and returns immediately with the CRITICAL log. The inner one catches generic exceptions from the pipeline run and result processing.

**New test:** `test_run_evolution_cycle_typeerror_in_pipeline_not_misclassified` — simulates a TypeError from inside `run_evolution_cycle()` and verifies it's handled as a generic error, not a configuration error.

### 2. ✅ Added `results_skipped` increment to codeless-result path

The codeless-result skip (missing `original_code`/`improved_code`) logged and `continue`d without incrementing `results_skipped`, while every other skip path (non-dict, bad fitness, collect failure, per-item exception) did increment it. This undercounted skips.

**Fix:** Added `self.service_stats["results_skipped"] += 1` to the codeless-result path.

**Updated test:** `test_run_evolution_cycle_skips_codeless_results` now asserts `results_skipped == 1`.

### 3. ℹ️ Pipeline call signature — already verified

Confirmed against real code:
- `EvolutionPipeline.__init__(config: dict[str, Any] | EvolutionConfig | None)` (line 95)
- `EvolutionConfig` has `seal_config: dict[str, Any]` field (line 68)
- `async def run_evolution_cycle(self, iterations: int = 1) -> list[dict[str, Any]]` (line 378)
- `EvolutionPipeline(config={"seal_config": seal_config})` correctly unpacks into `EvolutionConfig(**config)`

No change needed.

**Verification:**
- `ruff format --check` — pass
- `ruff check` — pass
- `pytest tests/unit/services/test_continuous_evolution_service.py` — 23/23 passed

PR #68, opened 2026-07-21
fix: validate_model now uses model_path instead of always baseline

SHA888:
## What

All 5 test suites in `ModelValidator` accepted a `model_path` parameter but silently ignored it, always constructing `OllamaProvider(model=self.baseline_model)`. This meant `validate_model()` could never actually validate a fine-tuned model — it always tested the baseline.

## Changes

- **`evoseal/fine_tuning/model_validator.py`**: Added `_resolve_model_for_validation(model_path)` helper that returns `model_path` when provided, falling back to `self.baseline_model` when `None`. All 5 test suites (`_test_basic_functionality`, `_test_code_generation_quality`, `_test_instruction_following`, `_test_safety_and_alignment`, `_test_performance_regression`) now call it instead of hardcoding the baseline.
- Removed the unused `TRANSFORMERS_AVAILABLE` / `torch` / `AutoModelForCausalLM` / `AutoTokenizer` import block (imported but never referenced anywhere in the file).
- **`tests/unit/fine_tuning/test_model_validator.py`** (new): 6 tests with mocked `OllamaProvider` verifying that `model_path` is forwarded correctly and baseline fallback works.

## Known limitation

`model_path` must be an Ollama-resolvable model tag (e.g. one registered via `ollama create`). Directory paths produced by `version_manager.register_version()` (which copies weights to a local dir) will surface an Ollama error rather than silently testing the wrong model. This is the expected dependency ordering per TODO.md — item 4 ("Implement real model deployment") will bridge the gap between filesystem paths and loadable Ollama models.

## Addresses

TODO.md item 3: "validate_model must serve the fine-tuned model, not baseline"

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now evaluates the specified fine-tuned model rather than always using the baseline.
  * Added stricter handling for model selection: unsupported/empty selections are rejected and fallback to the baseline occurs when no model is provided.
  * Validation results now reflect the selected model consistently.

* **Tests**
  * Added unit tests covering model resolution, suite forwarding, baseline fallback, and directory-path handling.

* **Documentation**
  * Updated the Phase 3 validation checklist in the project plan and added a known limitation for model tag requirements.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -107,7 +107,7 @@
 - [x] **1. Merge CLI wiring for `evoseal start evolution`** _(done 2026-07-21)_ — added `evoseal start evolution`, which constructs and runs `ContinuousEvolutionService` (the `api`/`worker` stubs are untouched); prints an explicit research-stage notice since the loop itself is still open per the gaps below. Along the way, fixed a pre-existing bug in `evoseal/cli/__init__.py` that clobbered the app's `--version` callback with a bare `lambda version: None`, breaking dispatch for every CLI subcommand invocation with arguments (the only prior CLI-dispatch test was a false positive that happened to pass regardless)
 - [x] **2. Wire daemon to real EvolutionPipeline** _(done 2026-07-21)_ — `continuous_evolution_service.py` `_run_evolution_cycle` now constructs and invokes `EvolutionPipeline.run_evolution_cycle()` instead of simulating; accepts optional `pipeline` param, lazy-inits if not injected
 - [x] **FIX: training call used nonexistent method** _(done 2026-07-19, commit fix/training-method-name-bug, on `main` as of 81156b3)_ — `continuous_evolution_service.py:234` called `training_manager.start_training()` which did not exist; changed to `run_training_cycle()` and fixed `validation_passed` → `validation_results.passed`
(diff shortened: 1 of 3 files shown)


SHA888:
## Review feedback addressed

### Point 1 — Empty-string `ValueError` crash
**Valid concern.** `_resolve_model_for_validation("")` did raise before the `try` block. Fixed by moving the resolve call inside the `try` so the existing `except Exception` handler catches it and returns a clean error dict. The only current caller (`training_manager.py:259`) constructs `model_save_path` from `self.output_dir / "fine_tuned_..."` which is never empty, but the defensive fix is still worth having.

### Point 2 — Removed `TRANSFORMERS_AVAILABLE` import block
**Not a problem.** Verified with `grep`: `TRANSFORMERS_AVAILABLE`, `torch`, `AutoModelForCausalLM`, and `AutoTokenizer` have zero references anywhere else in `model_validator.py`. The imports were genuinely dead. No change needed.

### Point 3 — Redundant `_resolve_model_for_validation` calls
**Valid.** Resolved once in `validate_model` and pass the resolved `model: str` to all 5 `_test_*` methods. Their signatures changed from `model_path: str | None` to `model: str` (internal methods, no external API impact). Eliminates 5 redundant calls.

### Also fixed (CodeRabbit inline comments)
- Added `assert len(captured_models) == 5` to the directory-path forwarding test.
- Fixed TODO.md co-evolution loop count: 4 done (items 1, 2, FIX, 3), not 5. Total: 24/48.

All checks pass: `ruff format`, `ruff check`, `pytest` (6/6).

PR #69, opened 2026-07-22
fix: mock network call in test_authentication_handling

SHA888:
## What

Mocks `CmdGit._run_git_command` in `test_authentication_handling` to simulate an authentication failure instead of making a real `git clone` call to GitHub.

## Why

The test previously called `repo.clone("https://github.com/private/nonexistent-repo.git")` against the real GitHub API. This caused:

- **Hangs in interactive TTYs** — a credential prompt blocks indefinitely when a credential helper is active
- **Stray directories on failure** — a `nonexistent-repo/` directory is left in the working tree
- **Flaky assertions** — error text varies by environment, causing intermittent failures

## How

- Added `from unittest.mock import patch`
- Wrapped the `clone` call with `patch.object(CmdGit, "_run_git_command", return_value=(False, stderr, ...))` to simulate the auth failure
- All original assertions on `GitError` message content are preserved

## Verification

```
295 files already formatted
All checks passed!

$ python -m pytest tests/version_control/test_advanced_operations.py -q
7 passed, 26 warnings in 12.23s
```

Addresses TODO.md item: *Fix `test_authentication_handling` making a real unmocked network call*

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -80,7 +80,7 @@
   - Verify the evolution loop can start (dependencies, permissions, Git state)
   - Flag budget/cost risks (no token limit configured, expensive model selected)
   - Surface risky configurations (e.g., immutable core protections disabled)
-- [ ] **Fix `test_authentication_handling` making a real unmocked network call** — `tests/version_control/test_advanced_operations.py:187` calls `repo.clone("https://github.com/private/nonexistent-repo.git")` against the real GitHub API instead of mocking it. Verified 2026-07-21: in an interactive TTY with a credential helper active, this hangs indefinitely on a live HTTPS username/password prompt instead of failing fast (blocked a real `git push`'s pre-push hook run this session); in batch mode it can also fail the `assert` when the returned error text doesn't match the expected substring list. Also leaves a stray `nonexistent-repo/` directory in the working tree on failure. Mock the clone failure instead of hitting the real network.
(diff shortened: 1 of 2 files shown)


SHA888:
## Review feedback addressed

Verified each point against the actual code. Here is the verdict:

### Point 1 — Mock target (`CmdGit._run_git_command`): ✅ Already correct
The reviewer questioned whether `clone()` actually calls `_run_git_command`. It does — `cmd_git.py:202` has:
```python
success, stdout, stderr = self._run_git_command(cmd, cwd=target_path.parent)
```
The mock target is correct. No change needed.

### Point 2 — Return value semantics `(False, "", ...)`: ✅ Already correct
The reviewer worried the first element might be `returncode: int` (where `False == 0` = success). Verified: `_run_git_command` returns `tuple[bool, str, str]` — the first element is `success: bool`, set at `git_interface.py:403` as `returncode == 0`. `False` unambiguously means failure. No change needed.

### Point 3 — `assert any(...)` substring match: ✅ Already correct
The mock stderr `"Permission denied (publickey)..."` triggers `clone()` to raise `SSHAuthenticationError("SSH authentication failed")` (subclass of `GitError`). The assert checks for `"authentication"` among others — present in `"ssh authentication failed"`. The test passes. Confirmed by the full test suite: `test_authentication_handling PASSED`.

### Point 4 — Missing closing `|` in TODO.md P1 row: ✅ Fixed
The P1 row was indeed missing its closing pipe character. Pushed fix in `17bcb29`.

PR #72, opened 2026-07-22
feat: implement real model deployment to Ollama

SHA888:
## What

Add `deploy_version()` method to `ModelVersionManager` that creates a Modelfile and runs `ollama create` so the serving layer (OllamaProvider) can actually load fine-tuned models. Previously, `register_version` only copied weights and updated a JSON registry — the model was never registered with Ollama.

## Changes

- **`evoseal/fine_tuning/version_manager.py`**: Added three new methods:
  - `deploy_version(version_id, ollama_model_name, base_url)` — public method that orchestrates deployment: creates Modelfile, runs `ollama create`, updates registry
  - `_create_modelfile(model_path)` — generates Modelfile content (handles both GGUF and HuggingFace-safetensors directories)
  - `_deploy_to_ollama(model_name, modelfile_path, base_url)` — runs `ollama create` via subprocess
  - Updated `register_version()` with `deploy=True` keyword arg to auto-deploy after registration
  - Added `deployment_distribution` to `get_version_statistics()`
- **`tests/unit/fine_tuning/test_version_manager_deploy.py`**: 14 unit tests covering success/failure paths, edge cases, and the register+deploy integration
- **`TODO.md`**: Checked off P2-4

## Why

Addresses TODO.md item P2-4: "Implement real model deployment". This is step 4 of 6 in closing the bidirectional co-evolution loop. Without this, the fine-tuned model weights sit in a version directory but are never loadable by the serving layer.

## Verification

```
ruff format --check .        ✅ 298 files formatted
ruff check evoseal/ tests/   ✅ All checks passed
pytest (full suite)           ✅ 512 passed, 23 deselected
```

## Summary by CodeRabbit

- **New Features**
  - Added optional automatic deployment of newly registered fine-tuned models to Ollama, with configurable deployment naming.
  - Added explicit deployment with recorded status, deployed model name, and timestamp.
  - Expanded deployment support for GGUF, adapter/LoRA-style, and directory-based model layouts, including conflict handling for overlapping deployment names.
  - Version statistics now include a deployment-status distribution.
- **Bug Fixes**
  - Improved deployment error reporting and refreshed the local installed-model cache after successful deployments.
- **Tests**
  - Added unit coverage for Modelfile creation, deployment success/failure paths, and statistics.
- **Documentation**
  - Updated project tracking to mark real model deployment as complete.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -170,7 +170,7 @@
 - [x] **2. Wire daemon to real EvolutionPipeline** _(done 2026-07-21)_ — `continuous_evolution_service.py` `_run_evolution_cycle` now constructs and invokes `EvolutionPipeline.run_evolution_cycle()` instead of simulating; accepts optional `pipeline` param, lazy-inits if not injected
 - [x] **FIX: training call used nonexistent method** _(done 2026-07-19, commit fix/training-method-name-bug, on `main` as of 81156b3)_ — `continuous_evolution_service.py:234` called `training_manager.start_training()` which did not exist; changed to `run_training_cycle()` and fixed `validation_passed` → `validation_results.passed`
 - [x] **3. validate_model must serve the fine-tuned model, not baseline** _(done 2026-07-21)_ — all 5 test suites now use `model_path` when provided via `_resolve_model_for_validation()`; removed dead `TRANSFORMERS_AVAILABLE` import block. **Known limitation:** `model_path` must be an Ollama-resolvable model tag — directory paths from `register_version()` will surface an Ollama error until item 4 (real deployment) lands.
-- [ ] **4. Implement real model deployment** — `version_manager.register_version` only copies weights + sets `current_version` in a JSON registry (no Modelfile / `ollama create` / symlink); need actual deployment so the serving layer can load the model
+- [x] **4. Implement real model deployment** — `version_manager.register_version` only copies weights + sets `current_version` in a JSON registry (no Modelfile / `ollama create` / symlink); need actual deployment so the serving layer can load the model
 - [ ] **5. Generation must consult the fine-tuning registry** — `version_manager.get_current_version()` has zero callers — the generator (`providers/local_models.py resolve_model`, which currently just reads raw installed Ollama tags) must consult the registry instead so the deployed fine-tuned model actually gets used
 - [ ] **6. Wire bidirectional_manager to orchestrate the full loop** — `bidirectional_manager.py`'s docstring promises evolve → collect → train → deploy → repeat, but the class only implements reporting/statistics methods (`get_evolution_status`, `get_evolution_history`, `generate_evolution_report`); no method actually drives the sequence end-to-end
 - [ ] **Add end-to-end bidirectional loop test** — exercise the full cycle: collect → train → validate → deploy → regenerate; no such test exists today (write once steps 1–6 land)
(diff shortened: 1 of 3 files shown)


SHA888:
## Review feedback addressed

### 1. Unquoted paths in `_create_modelfile` ✅ Fixed
Added `_validate_model_path()` which rejects paths containing whitespace (spaces, tabs, newlines) before they reach the Modelfile template. `_create_modelfile` calls it on both the directory and the selected GGUF file. `deploy_version` catches the `ValueError` and returns a clear error dict. This is a fail-fast approach rather than quoting — Ollama's Modelfile parser behavior with quoted paths isn't well-documented, so rejecting bad paths up-front is safer than guessing at quoting semantics.

### 2. Aliasing dependency in `register_version` ✅ Documented (no code change)
Verified: `get_version_info` (line ~228) iterates `self.registry["versions"]` and returns the matching dict by reference — mutations in `deploy_version` propagate correctly today. Added a NOTE comment on the `register_version` return path documenting this implicit coupling so a future change to `get_version_info` (e.g., returning a copy) won't silently break the return value.

### 3. Unsanitized `ollama_model_name` ✅ Fixed
`deploy_version` now sanitizes the model name: lowercased, non-`[a-z0-9._-]` characters replaced with hyphens, leading/trailing punctuation stripped. Falls back to `model-{version_id}` if the result is empty. Logs the transformation when it changes.

### 4. Tautological assertion ✅ Fixed
Simplified `assert "not found" in result["error"].lower() or "not found" in result["error"]` → `assert "not found" in result["error"].lower()`.

### 5. Redundant local `import json as _json` ✅ Fixed
Removed the inline import; uses module-scope `json` instead.

### Tests added
- `TestPathValidation`: 4 tests — spaces in dir name, spaces in GGUF filename, clean path succeeds, integration test for `deploy_version` rejecting whitespace paths.
- `TestModelNameSanitization`: 3 tests — uppercase lowered, spaces replaced, special chars replaced.

**Verification:**
```
ruff format --check .        ✅ 298 files formatted
ruff check evoseal/ tests/   ✅ All checks passed
pytest                       ✅ 521 passed, 23 deselected
```

SHA888:
Addressed all 5 review points in commit dd8364f. Verification against actual code:

**1. Supersede loop ordering** ✅ Fixed — moved the supersede loop to *after* `_deploy_to_ollama` returns success. Added a `deployment_status == "deployed"` guard so re-deploying the same version under the same name doesn't self-supersede. Three new tests confirm failed `ollama create` and failed `_create_modelfile` do NOT mark the old version as superseded.

**2. `base_model` not validated** ✅ Fixed — added `ModelVersionManager._validate_model_path(Path(base_model))` in the adapter branch of `_create_modelfile`. A whitespace-containing `base_model_name_or_path` from `adapter_config.json` now raises `ValueError` with the same message the other branches get. New test covers this.

**3. `deployment_error` not persisted** ✅ Fixed — added `self._save_registry()` after setting `deployment_error` in the "deploy requested but no model_path" branch. New test reloads from disk and verifies the error persists.

**4. `.get(key, default)` with `None` value** ✅ Defensive fix — changed `version_info.get("version_name", version_id)` to `version_info.get("version_name") or version_id` and added an `ollama_model_name = ollama_model_name or version_id` guard before `.lower()`. Practically impossible today (register_version always generates a string), but now safe against future changes.

**5. Reference-return caveat** ℹ️ Already documented in-code. `get_version_info` returns by reference (confirmed: iterates `self.registry["versions"]` and returns the matching entry directly). No change needed.

All 527 tests pass. 6 new regression tests added.

SHA888:

**1.  registry mutation fragility** — Replaced the  call with a direct loop over . Mutations to , , , etc. now always land in the registry list regardless of whether  returns by reference or by copy. The defensive re-fetch in  is kept as a belt-and-suspenders measure.

**2. Non- exceptions from deploy masking registration** — Wrapped the  call inside  in a narrow . A deploy-time crash (e.g. from , , dict/JSON handling) now logs a warning and returns the version info with deployment failure fields, instead of propagating to the outer  which would report the whole call as an error despite the version already being persisted.

**3.  silent degradation on bad adapter config** — When  exists but  is missing, empty, or the file is corrupt/unreadable,  now raises  instead of falling through to . This is consistent with 's error-on-ambiguity approach and prevents confusing downstream failures from ollama.

**Tests added:**
-  — config without 
-  — empty string value
-  — corrupt JSON
-  —  from  does not mask registration

All 531 tests pass, ruff clean.

SHA888:

**1. deploy_version registry mutation fragility** — Replaced the `get_version_info()` call with a direct loop over `self.registry["versions"]`. Mutations now always land in the registry list regardless of whether `get_version_info()` returns by reference or by copy.

**2. Non-ValueError exceptions from deploy masking registration** — Wrapped the `deploy_version()` call inside `register_version` in a narrow `try/except Exception`. A deploy-time crash now logs a warning and returns version info with deployment failure fields, instead of propagating to the outer except which would report the whole call as an error despite the version already being persisted.

**3. _create_modelfile silent degradation on bad adapter config** — When `adapter_config.json` exists but `base_model_name_or_path` is missing, empty, or the file is corrupt, `_create_modelfile` now raises `ValueError` instead of falling through to a bare `FROM <dir>`.

**Tests added:**
- `test_adapter_missing_base_model_raises`
- `test_adapter_empty_base_model_raises`
- `test_adapter_unparseable_config_raises`
- `test_unexpected_deploy_error_still_returns_version`

All 531 tests pass, ruff clean.

PR #84, opened 2026-07-23
fix: use pydantic.ValidationInfo instead of undefined FieldValidationInfo

SHA888:
## What

Replace the undefined `FieldValidationInfo` type annotation in `Experiment.ensure_timezone_aware` with the correct `pydantic.ValidationInfo`, and add it to the import block.

## Why

`models/experiment.py:256-260` referenced `FieldValidationInfo` — a name that is not a pydantic export and was never imported. It only survived because `from __future__ import annotations` defers evaluation. Under `typing.get_type_hints()`, strict mypy/pyright, or Sphinx autodoc this would break.

Addresses TODO.md item: **`models/experiment.py:256-260` references an undefined name `FieldValidationInfo`**

## How

- Added `ValidationInfo` to the `from pydantic import ...` block
- Changed the type annotation from `FieldValidationInfo` to `ValidationInfo`

## Verification

- `ruff format --check .` ✅ (304 files clean)
- `pytest tests/unit/models/` — 15/15 passed ✅
- `pytest` (full suite via pre-commit) — 576 passed ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation typing for experiment date/time handling to better align with the current validation framework.
* **Documentation**
  * Updated the project checklist to reflect completion of the related validation task.
  * Refreshed many documentation examples for consistency (quoting, whitespace, and code-block formatting) across API reference, guides, architecture, safety, events, and integration materials.


SHA888:
## Review feedback addressed

Verified all three points against the actual repo state:

**Point 1 — TODO.md checkbox vs code fix**: The code fix *is* present in this PR. The diff includes both:
- `ValidationInfo` added to the `from pydantic import (...)` block in `evoseal/models/experiment.py` (line 20)
- Type annotation changed from `FieldValidationInfo` → `ValidationInfo` (line 259)
- `TODO.md` checkbox flipped `[ ]` → `[x]` for the corresponding item

The reviewer's concern was based on incomplete diff visibility — the fix and the checkbox change are both here and consistent.

**Point 2 — Truncated diff in `evolution_pipeline_safety_integration.md`**: Confirmed as diff-display truncation only. All 10 code blocks in the file are properly closed; the `safety_config={...}` construction has balanced braces (29 opens, 29 closes). No integrity issue.

**Point 3 — Formatting-only changes**: Agreed, no semantic issues.

**No changes needed.** All CI checks green.

PR #88, opened 2026-07-24
fix: validate from_yaml returns a dict, not None/scalar/list

SHA888:
## What

`SystemConfig.from_yaml()` passed `yaml.safe_load()` output directly to `cls()` without type-checking. An empty YAML file yields `None`, a scalar yields `str`/`int`, and a list yields a `list` — all cause an opaque `TypeError` on the first `.get()` or `.validate()` call.

## How

Added a type check after `yaml.safe_load()`: if the result is not a `dict`, raise `ValueError` with a clear message identifying the file path and the actual type returned.

## Tests

Added 3 regression tests:
- `test_from_yaml_empty_file_raises` — empty file → `None` → `ValueError`
- `test_from_yaml_scalar_raises` — scalar document → `str` → `ValueError`
- `test_from_yaml_list_raises` — list document → `list` → `ValueError`

All 6 tests in `tests/unit/models/test_system_config.py` pass. Ruff format + check clean.

## Addresses

TODO.md P2 item: `models/system_config.py:33-39 from_yaml doesn't validate the loaded YAML is a dict`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved YAML configuration loading by rejecting empty files and YAML values that are not mappings.
  * Added clearer validation errors identifying invalid configuration input.

* **Tests**
  * Added coverage for empty, scalar, and list-based YAML configuration files.

* **Documentation**
  * Updated the project task tracker to reflect completion of the configuration validation fix.


SHA888:
Addressed review feedback:

1. **`os`/`tempfile`/`yaml` imports** — verified against the actual file; all three are already imported at the top of `test_system_config.py` (lines 1–6). The diff didn't show them because they were untouched, but they exist. No change needed.

2. **`yaml.YAMLError` guard** — agreed this is a real gap. Added a follow-up TODO item since the original item is marked done and this is out of scope for the current PR. All 603 tests pass, ruff clean.

PR #89, opened 2026-07-24
test: add end-to-end bidirectional loop integration test

SHA888:
## What

Adds an end-to-end integration test for the bidirectional evolution loop that exercises the full cycle: **collect → train → validate → deploy → record**.

## Why

The existing tests in `test_bidirectional_manager.py` mock `TrainingManager` entirely (unit-level). This integration test wires **real** component instances together — `BidirectionalEvolutionManager`, `TrainingManager`, `EvolutionDataCollector`, `ModelVersionManager` — with mocks only at the external boundary (GPU fine-tuning, Ollama validation/deploy). This verifies the real orchestration wiring works end-to-end.

## Test cases (5)

1. **Full success cycle** — readiness → train → validate → deploy → stats updated
2. **Readiness skip** — insufficient training candidates records a skipped cycle
3. **Training failure propagation** — OOM-style failure surfaces as `success=False, phase=training`
4. **Validation failure skips deploy** — low score means deploy is skipped, no improvement counted
5. **Two sequential cycles** — stats and history accumulate correctly across back-to-back runs

## Addresses

`TODO.md` P2 item: "Add end-to-end bidirectional loop test"

## Verification

```
All checks passed!

$ python -m pytest tests/integration/test_bidirectional_loop_e2e.py -v
5 passed

$ python -m pytest (full suite)
600 passed, 23 deselected
```

## Summary by CodeRabbit

* **New Features**
  * Added the ability to fetch recent *successful* evolution results from the last N days, combining in-memory and on-disk data with deduplication.

* **Bug Fixes**
  * Improved the bidirectional co-evolution training flow by addressing latent collect→train path issues uncovered during real end-to-end execution.

* **Tests**
  * Added an end-to-end integration test covering the full bidirectional workflow (readiness → training → validation → deploy), including skipped and failure scenarios.
  * Verified cycle IDs and stats/history across repeated successful runs.

* **Documentation**
  * Updated the tracking checklist and totals in the project TODO file for this completed end-to-end test work.


SHA888:
## Review feedback addressed

Verified all 6 points against the actual source. Here is the verdict and action on each:

### Applied (3 points)

1. **Duplicate `original_get_stats`/`patched_get_statistics` boilerplate (×5)** — Extracted into a shared `_patch_training_candidates(dc, count)` helper that returns a context manager. Reduces 43 lines of copy-paste to a single reusable function.

2. **`patched_get_statistics` mutation risk** — Added a defensive `dict(original())` copy in the helper. Note: `get_statistics()` already returns a fresh dict literal (not a reference to internal state), so this is strictly a belt-and-suspenders measure against future refactors — no corruption was possible today.

3. **`last_training_time = None` whitebox hack** — Added a clarifying comment explaining that `last_training_time` is a public attribute (documented in `get_training_status`) and no `force=True` API exists on the readiness check. Directly clearing it is the simplest correct approach given the current API surface.

### Verified correct — no change needed (3 points)

4. **`cycle_id == 0` and `== 1` hardcoded** — The reviewer speculated these might be UUIDs or timestamps. They are not. `_record_cycle()` in `bidirectional_manager.py` returns `len(self.evolution_history) - 1` — a 0-based sequential index. The assertions are correct as written.

5. **`version_id` collision risk (flaky)** — `_generate_version_id()` in `version_manager.py` uses `datetime.now().isoformat()` (microsecond precision) + an MD5 hash of the full `training_results` dict. Two back-to-back calls will always produce different timestamps. Collision risk is negligible; no mock/counter needed.

6. **`fine_tuner`/`validator` as `patch.object` targets** — Both are public attributes set in `TrainingManager.__init__` (`self.fine_tuner = ...`, `self.validator = ...`), not private internals. They are stable public API suitable for `patch.object`.

### Verification

```
All checks passed!

$ python -m pytest tests/integration/test_bidirectional_loop_e2e.py -v
5 passed

$ python -m pytest (full suite)
600 passed, 23 deselected
```

SHA888:
## Review feedback analysis

Verified all 4 points from the reviewer and 2 CodeRabbit inline comments against the actual source. **No code changes needed.**

### Reviewer points

| # | Concern | Verdict | Reason |
|---|---------|---------|--------|
| 1 | `fine_tuner`/`validator` attrs might be private/differently named | **Incorrect** | `TrainingManager.__init__` (lines 56–58) sets `self.fine_tuner` and `self.validator` as **public** attributes. `ModelFineTuner.fine_tune_model` (line 280) and `ModelValidator.validate_model` (line 58) both exist with those exact names. All 5 tests pass — no `AttributeError` at setup. |
| 2 | `dict(original())` assumes plain dict return from `get_statistics()` | **Incorrect** | `get_statistics()` is annotated `-> dict[str, Any]` (line 252) and returns a plain dict literal. `dict()` on a dict is a safe defensive copy. |
| 3 | `last_training_time = None` couples test to internal field name | **Marginal, already handled** | The attribute is public (no `_` prefix), exposed through `get_training_status()` (line 300). There is no public reset API. The test already has an inline comment explaining the rationale. If the field were renamed, the test would fail with `AttributeError` — not silently. |
| 4 | 750 sequential `collect_result` calls might slow CI | **Not a problem** | All 5 tests complete in **2.16 seconds** total. The seeding exercises the real collection path (pattern tracking, callbacks, persistence) as documented in the docstring. |

### CodeRabbit inline comments

- **`prepare_training_data` needs `AsyncMock`** — Incorrect. `patch.object` auto-detects `async def` methods and creates `AsyncMock` by default (confirmed empirically). The tests pass (5/5).
- **Create mocked model artifact directories** — Not a bug. `ModelVersionManager.register_version()` guards the `shutil.copytree` call with `if model_path and Path(model_path).exists()` (line 138), so missing paths are gracefully skipped (version registers as "registered" instead of "stored"). Tests don't assert "stored" status.

PR #90, opened 2026-07-25
fix: correct type annotation for temp_file suffix parameter

SHA888:
## What

Fixed type-annotation mismatch in `evoseal/utils/testing/environment.py:157`:
`suffix: str = None` → `suffix: Optional[str] = None`.

`Optional` was already imported in the module; no new imports needed.

## Why

`None` is not a valid `str` default. The previous annotation was a type mismatch
that would fail under strict type checking (mypy/pyright). The function already
handles `None` correctly via `suffix or ""` on line 169, so this is annotation-only
— no behavioral change.

## Addresses

TODO.md item: "type-annotation mismatch (should be `Optional[str]`)" under
🟢 P3 — Low-Priority / Hygiene Issues.

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/testing/ tests/unit/utils/test_testing_utils.py -q` → 37 passed ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected the type annotation for an optional temporary-file suffix, improving compatibility with static type checking without changing runtime behavior.

* **Documentation**
  * Updated project tracking to reflect completion of a low-priority code hygiene item.


SHA888:
Re: `Optional` import — verified against the actual file. Line 15 of `evoseal/utils/testing/environment.py` already imports `Optional` from `typing`:

```python
from typing import Any, Dict, Optional, Union
```

This import existed before this PR. No change needed.

PR #93, opened 2026-07-25
fix: add HuggingFace dataset export to TrainingDataBuilder

SHA888:
## What

Added `format_type="huggingface"` to `TrainingDataBuilder.save_training_data()` so the collect→train pipeline produces a format that `ModelFineTuner.fine_tune_model()` can actually load.

## Why

`ModelFineTuner.fine_tune_model()` (non-fallback path) calls `datasets.load_from_disk()` which expects an HF datasets Arrow directory. `TrainingDataBuilder.save_training_data()` only produced `alpaca`/`chat`/`jsonl` formats (all plain JSON files). This format mismatch would crash fine-tuning whenever `transformers`/`peft` are installed and the real training path runs.

This was the last remaining gap in the bidirectional co-evolution loop (P2 item in TODO.md).

## Changes

- Added `_save_huggingface_format()` to `TrainingDataBuilder` — uses `Dataset.from_list()` + `save_to_disk()` for both train and val splits
- Updated `TrainingManager.prepare_training_data()` to use `format_type="huggingface"` and point `data_file` to the train split directory
- Added 3 unit tests verifying the saved datasets load correctly via `load_from_disk` and contain the expected columns

## Addresses

TODO.md item: `ModelFineTuner.fine_tune_model()` format mismatch (P2 — Close the bidirectional co-evolution loop)

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/evolution/test_training_data_builder.py -v` → 3 passed ✅
- `pytest tests/unit/fine_tuning/ -v` → 57 passed ✅
- `pytest tests/integration/test_bidirectional_loop_e2e.py -v` → 5 passed ✅

## Summary by CodeRabbit

* **New Features**
  * Training data export now supports a Hugging Face–compatible format, including separate `train/` and `val/` artifacts, Alpaca-style columns, and saved metadata.
* **Bug Fixes**
  * Fine-tuning now prefers the Hugging Face export path (with safe fallback) and ensures loaded datasets are tokenized consistently when `input_ids` is missing.
* **Tests**
  * Added unit tests for Hugging Face export/loading and for conditional tokenization.
  * Updated integration tests to skip when the optional Hugging Face datasets dependency is unavailable.
* **Chores**
  * Updated test extras to include the required `datasets` version.


SHA888:
Addressed review feedback. Here's the verdict on each point after verifying against the actual code and runtime:

### Point 1 — Empty split crash: **Fixed** ✅

Confirmed as a real bug, and worse than described. `Dataset.from_list([])` produces a columnless dataset, and `load_from_disk` on it actually crashes with `IndexError: list index out of range` (not just a schema mismatch). With 1 example and default `split_ratio=0.8`, `split_idx=0` → empty train → crash.

**Fix (two layers):**
- `save_training_data`: ensure `split_idx >= 1` when there are examples, so train is never empty.
- `_save_huggingface_format`: skip `save_to_disk` for empty splits as a safety net, returning only keys for non-empty splits.

Added a single-example edge-case test (`test_save_huggingface_single_example`) that would have crashed before the fix.

### Point 2 — val directory discarded: **Acknowledged, not a regression** ℹ️

Confirmed as a caveat. `prepare_training_data` now returns `data_dir / "train"` and the val HF directory is unused. However, this is not a regression:
- The old code used `format_type="jsonl"` and pointed `data_file` at `data_dir` (containing `train.jsonl` / `val.jsonl`).
- `fine_tune_model` calls `load_from_disk(training_data_path)` on that path — which would have crashed on a directory of JSONL files anyway.
- The val split was never consumed — `fine_tune_model` has no eval/validation logic.

### Point 3 — Pre-existing directories: **Not an issue** ❌

Tested `Dataset.save_to_disk()` on an existing directory — it silently overwrites without error. No fix needed.

SHA888:
Addressed all three review points. Here's the assessment against actual repo state and what changed:

### Point 1: `datasets` as hard dependency for all training-data prep
**Reviewer was correct.** `_save_huggingface_format` raises `RuntimeError` if `datasets` is missing, and `prepare_training_data` called it unconditionally — crashing even when the fallback fine-tuning path (which never loads HF data) would have worked fine.

**Fix:** Wrapped the `huggingface\format` call in a try/except `RuntimeError`. When `datasets` is missing, logs a warning and falls back to `format_type="jsonl"`. The fallback training path gets usable data; the real training path would still fail at `load_from_disk` time if someone tries to run it without `datasets`, which is the correct place for that failure.

### Point 2: Val split path dropped from return dict
**Partially valid.** The val split was already never consumed downstream — `fine_tune_model`'s `Trainer` receives no `eval_dataset`, and the pre-PR code also didn't pass val. However, writing data to disk and not returning the path is wasteful.

**Fix:** Added `val_data_file` to the result dict when the val split directory exists. The downstream `Trainer` still doesn't use it (pre-existing gap), but the path is now available for future use.

### Point 3: `datasets` not declared as a dependency
**Reviewer was correct.** `datasets>=2.10.0` was only under `[project.optional-dependencies] benchmarks`, not in a group that training-path code would install.

**Fix:** Added `datasets>=2.10.0` to the `test` optional dependencies in `pyproject.toml`, alongside `transformers` (which has the same status — required for the real training path, optional for the fallback). Not added to core `dependencies` because the fallback path exists precisely for environments without these heavy ML packages.

SHA888:
Traced the full call chain for the real (non-fallback) training path and I don't think this closes the gap yet — it moves the failure point rather than removing it.

**Call chain:** `TrainingManager.run_training_cycle()` → `prepare_training_data()` → `self.fine_tuner.fine_tune_model(training_data_path=data_prep_results["data_file"])` (`training_manager.py:250-253`).

In `ModelFineTuner.fine_tune_model()` (`model_fine_tuner.py:346-388`), the loaded dataset goes straight into `Trainer`:

```python
dataset = load_from_disk(training_data_path)
...
data_collator = DataCollatorForLanguageModeling(tokenizer=self.tokenizer, mlm=False)
trainer = Trainer(model=self.peft_model, args=training_args, train_dataset=dataset, data_collator=data_collator)
trainer.train()
```

`DataCollatorForLanguageModeling` requires each example to already be tokenized (`input_ids`/`attention_mask`). But `TrainingDataBuilder._save_huggingface_format()` builds the dataset from `TrainingExample.to_alpaca_format()`, which is raw strings:

```python
{"instruction": ..., "input": ..., "output": ...}
```

No tokenization happens in the new `_save_huggingface_format()`, and `fine_tune_model()` doesn't tokenize either. There's a separate method, `ModelFineTuner.prepare_training_data()` (`model_fine_tuner.py:192-278`), that *does* do this correctly — it loads examples, formats them, and runs `dataset.map(tokenize_function, batched=True, remove_columns=...)` before saving — but it's never called from `run_training_cycle()`. It's dead code for this pipeline both before and after this PR.

So this PR fixes the `load_from_disk()`-needs-an-Arrow-directory problem (real), but the dataset schema still doesn't match what `Trainer` + `DataCollatorForLanguageModeling` need. With `transformers`/`peft` installed and a real model initialized, `fine_tune_model()` would now get past `load_from_disk()` only to crash inside `trainer.train()` (missing `input_ids`) — still the format-mismatch bug the PR is meant to close, just one layer downstream.

**Why the test suite didn't catch it:** there's no test file for `model_fine_tuner.py` anywhere in the repo, and `test_bidirectional_loop_e2e.py` explicitly mocks GPU fine-tuning per its own docstring, so `fine_tuner.fine_tune_model()`'s real body is never actually executed by CI. The 3 new unit tests validate that `TrainingDataBuilder` produces a loadable `Dataset`, but never feed that dataset into `ModelFineTuner`.

**Suggested fix:** move tokenization into `fine_tune_model()` itself (it already owns `self.tokenizer`) — map the loaded dataset through the same tokenize step currently stranded in `prepare_training_data()` before constructing `Trainer`. That also makes `prepare_training_data()`'s dead code path either used or removable.

Happy to be wrong here — let me know if there's a tokenization step I'm missing elsewhere in the chain.

SHA888:
Closed the tokenization gap from the earlier comment.

**Fix:** added `ModelFineTuner._tokenize_if_needed()`, called immediately after `load_from_disk()` in `fine_tune_model()`. It checks for `input_ids` in the loaded dataset's columns — if absent (the case for the new HF export's raw `instruction`/`input`/`output` columns), it formats each example with the same `<s>[INST] ... [/INST] ...</s>` template `prepare_training_data()` already uses and tokenizes via `self.tokenizer`. If `input_ids` is already present (e.g. a dataset that went through the older, still-dead `prepare_training_data()` path), it's a no-op — so both paths now produce a dataset `Trainer` + `DataCollatorForLanguageModeling` can actually consume.

**Verification:**
- Added `tests/unit/fine_tuning/test_model_fine_tuner.py` (3 tests): tokenization is skipped when `input_ids` already exists, an Alpaca-format dataset gets tokenized and its raw string columns removed, and `max_length` is correctly threaded through to the tokenizer call.
- Ran the full relevant suite (`tests/unit/fine_tuning/`, `tests/unit/evolution/`, `tests/integration/test_bidirectional_loop_e2e.py`) — 69 passed.
- Full pre-push suite: 624 passed, 23 deselected.
- `ruff format` / `ruff check` clean.

Also updated the TODO.md entry for the original format-mismatch bug to note this follow-up, since its wording claimed the gap was fully closed when it wasn't yet.

PR #95, opened 2026-07-26
fix: prevent path traversal in git file read/write helpers

SHA888:
## What

Added path traversal protection to `CmdGit.get_file_content()` and `CmdGit.write_file_content()`.

## Why

Both methods constructed file paths via `self.repo_path / file_path` with no containment check. `pathlib` resolves an absolute RHS by discarding the LHS (`Path('/repo') / '/etc/passwd'` → `/etc/passwd`), and `..` segments aren't normalized. Any caller passing an absolute or `..`-containing `file_path` (e.g. a model-generated patch) could read or write arbitrary files outside the repository.

## Changes

- Both methods now resolve the constructed path and verify it stays within `repo_path` via `is_relative_to()`, raising `GitError` on traversal
- Added 4 tests: dot-dot traversal blocked on read, absolute path blocked on write, nested traversal blocked, and legitimate subdirectory ops still work

## Addresses

TODO.md item: "Path traversal in git-file read/write helpers" (P1 — Safety & Correctness Bugs)

## Verification

- `ruff format --check .` ✅
- `pytest tests/version_control/test_advanced_operations.py -v` → 10 passed ✅

## Summary by CodeRabbit

* **Bug Fixes**
  * Blocked file read and write operations that attempt to escape the repository directory (including traversal via symlinks).
  * Continued to allow valid operations within normal nested subdirectories.

* **Tests**
  * Added cases covering relative traversal, absolute paths, nested traversal variants, and symlink-based traversal.

* **Documentation**
  * Updated safety/correctness tracking and completion counts to reflect the fix.


SHA888:
## Review feedback — addressed where needed

Went through each point against the actual repo state:

### 1. `is_relative_to()` requires Python ≥3.9 — **Non-issue** ✅
`pyproject.toml` specifies `requires-python = ">=3.11"`. `is_relative_to()` was added in 3.9, so both call sites are safe. No change needed.

### 2. `resolve()` not wrapped in try/except — **Already handled** ✅
Both `get_file_content` and `write_file_content` have an outer `try/except Exception` block that wraps the entire function body. Any `OSError` from a symlink loop or permission error during `resolve()` is caught by that outer handler and re-raised as `GitError("Failed to get/write file content: ...")`. The error message won't say "escapes repository root" (it'll be the raw `OSError` text), but no raw `OSError` leaks to callers. Adding a dedicated try/except around `resolve()` would be over-engineering given the existing guard.

### 3. Scope check — other methods with the same pattern — **Valid observation, low severity** ℹ️
Two other call sites use `self.repo_path / file_path`:
- **`get_file_history` (line 1462)**: `(self.repo_path / file_path).exists()` — existence check only, no content read/write. Symlink traversal could theoretically probe for arbitrary file existence (info disclosure), but the risk is low since the method returns git log data, not file contents.
- **`find_file_references` (line 1615)**: `file_path.relative_to(self.repo_path)` — this would raise `ValueError` (not `GitError`) on traversal, so it fails closed, just with a less clean error. Symlink-based traversal isn't guarded but the method only returns reference metadata, not file data.

Neither is as critical as the read/write paths. If you want them hardened too, that's a follow-up task — not blocking for this PR.

### 4. TODO.md "4 tests" vs 3 test functions — **Doc nit, fixing now** 🔧
The diff adds 3 test functions, not 4. Updating the changelog entry.

SHA888:
Addressed review feedback — here's the verdict on each point:

**1. `Path.is_relative_to()` requires Python 3.9+ → No change needed.**
`pyproject.toml` declares `requires-python = ">=3.11"`. `is_relative_to()` was added in 3.9. No 3.8 risk.

**2. Missing symlink test → Fixed.**
Valid point — `resolve()` does follow symlinks (so the fix was correct), but there was no test proving it. Added `test_symlink_traversal_blocked` which creates a symlink inside the repo pointing to a directory outside it, and verifies both `get_file_content` and `write_file_content` reject the traversal. All 608 tests pass.

**3. Error message echoes raw `file_path` → Acknowledged, no change.**
The reviewer flagged this as "not a real vuln" and "worth being mindful of." The error is a `GitError` raised and caught internally — it doesn't reach logs or API responses in this code path. Changing it now would be speculative; if the error surface changes later, this is worth revisiting.

SHA888:
### Re: review feedback

**TOCTOU race** — Acknowledged. You're right that there's a residual window between `resolve()` and the I/O call. The code does the best pathlib can do: `resolve()` is called once and the same resolved path object is used for the actual `read_text()`/`write_text()`, so we're not re-traversing the original path. The remaining attack requires swapping a directory component with a symlink between `resolve()` and `open()`, which needs write access inside the repo's own tree mid-operation. Agreed this is a known limitation of the pathlib-only approach; fd-based `O_NOFOLLOW` checks would be the proper fix if this becomes a concern. Noting as residual risk.

**`is_relative_to()` and Python version** — Confirmed not an issue. `pyproject.toml` has `requires-python = ">=3.11"`. `is_relative_to()` was added in 3.9. Safe to use.

PR #97, opened 2026-07-26
fix: move signal handler registration from __init__ to start()

SHA888:
## What

Move signal handler registration out of `ContinuousEvolutionService.__init__` and into `start()`.

## Why

The previous code called `signal.signal()` from `__init__` (line 103), which has three problems:

1. **Thread restriction**: `signal.signal()` only works on the main thread — constructing the service off the main thread raises `ValueError`
2. **Missing event loop**: The handler used bare `asyncio.create_task(self.shutdown())`, but no event loop exists during `__init__` — this silently creates an unawaited coroutine that's immediately discarded
3. **Premature clobbering**: Signal handlers are clobbered on mere construction, not on `start()`, so an embedding application that creates the service but doesn't start it immediately loses its own SIGINT/SIGTERM handlers

## How

- Move `_setup_signal_handlers()` call from `__init__` to `start()` (after `asyncio.get_running_loop()` is available)
- Use `loop.call_soon_threadsafe(asyncio.ensure_future, self.shutdown())` instead of bare `asyncio.create_task()` — schedules the coroutine on the event loop from the signal context safely
- Save original signal handlers on install, restore them in `_cleanup()` on shutdown
- Add `_original_sigint` / `_original_sigterm` instance attributes

## Verification

```
$ ruff format --check evoseal/services/continuous_evolution_service.py
1 file already formatted

$ ruff check evoseal/services/continuous_evolution_service.py
All checks passed!

$ pytest tests/unit/services/test_continuous_evolution_service.py -q
26 passed, 17 warnings
```

Addresses TODO.md P1 item: `continuous_evolution_service.py:107-115` registers process-wide signal handlers from `__init__`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graceful shutdown handling for service interrupts with idempotent shutdown behavior.
  * Signal handlers are now installed only when the service starts and are restored during cleanup, preventing interference with existing application signal behavior.

* **Documentation**
  * Updated tracking information to reflect the completed fix and incremented the related completion counters.


SHA888:
Addressed review feedback. Here is the point-by-point verification against the actual code on this branch:

**Point 1 — Double `start()` clobbering originals:** Already handled. `start()` has an `if self.is_running: return` guard (line 134) that fires before `_setup_signal_handlers()` (line 144), and `self.is_running = True` is set at line 139. A second call while the first is in progress hits the guard. After `_cleanup()` runs (in `finally`), `_original_*` are `None`, so a subsequent `start()` correctly captures the real originals. No change needed.

**Point 2 — Concurrent `shutdown()` re-entrancy:** Valid. `shutdown()` had no guard — rapid signals schedule multiple coroutines via `call_soon_threadsafe`, and `generate_service_report()` has await points, so concurrent coroutines could race on file I/O. **Fixed** by adding a `_shutting_down` flag checked at the top of `shutdown()` (commit `ed1df6c`).

**Point 3 — `Any` import:** Already present at line 14 (`from typing import Any`). No change needed.

**Point 4 — `_cleanup()` via `finally`:** Confirmed. Line 154: `finally: await self._cleanup()` wraps the entire try block in `start()`. Python `finally` covers all exit paths including `CancelledError`. No change needed.

SHA888:
Addressed review feedback with a follow-up commit:

**Point 1 (signal handler setup before `try`)** — Fixed. Moved `self._setup_signal_handlers()` inside the `try` block so that if `signal.signal()` raises (e.g. off the main thread), the `finally` block still resets `is_running` and runs `_cleanup()`.

**Point 2 (unretrieved coroutine on closed loop)** — Fixed. Added a `loop.is_closed()` guard before `call_soon_threadsafe`, plus a `RuntimeError` catch around it, so a shutdown coroutine is never scheduled on a dead loop.

**Point 3 (cleanup/restore reachability)** — Verified: the `try/except/finally` wrapping `_run_service_loop()` already ensures `_cleanup()` (and thus `_restore_signal_handlers()`) runs on every exit path. No change needed.

Verification: `ruff format --check .` ✅ · `ruff check evoseal/ tests/` ✅ · `pytest` 604 passed ✅

PR #101, opened 2026-07-27
fix: use atomic write for version registry to prevent data loss on crash

SHA888:
## What

`ModelVersionManager._save_registry()` now writes to a temporary file first, then performs an atomic `os.replace()`, so a crash or kill mid-write never leaves the registry file truncated.

## Why

Previously, `_save_registry()` wrote directly via `open(..., "w")` + `json.dump()`. A crash mid-write would truncate the file, and `_load_registry()` silently resets to an empty registry on parse failure — losing all version history with no error surfaced.

This is the **atomic write** item from TODO.md P2.

## Changes

- `evoseal/fine_tuning/version_manager.py`: `_save_registry()` now uses `tempfile.mkstemp()` + `os.replace()` for atomic writes. Temp files are cleaned up on failure. Parent directories are created if needed.
- `tests/unit/fine_tuning/test_version_manager_atomic_write.py`: 7 tests covering round-trip, no orphan temp files, write-failure-preserves-original, and first-save-to-nonexistent-dir.
- `TODO.md`: Flipped checkbox, updated tracking table (P2: 15→16).

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/fine_tuning/ -v` ✅ (64 passed)
- Full `pytest tests/` ✅ (611 passed, 23 deselected)

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved registry saving to use an atomic temp-file write and replace, preventing truncated or partially written registry files.
  - Preserved existing registry file permissions across saves, and set safe default permissions for new registries.
  - Added/strengthened cleanup to ensure temporary files don’t remain after save success or simulated write failures.
  - Ensures missing directories are created when the registry location doesn’t yet exist.
- **Documentation**
  - Updated the tracking status to mark the registry atomic-write issue as completed.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -308,7 +308,7 @@
 | 🟠 P1    | 24    | 11   | Original safety/integration items done; +12 high-priority bugs from 2026-07-22 review |
 | 🟡 P2    | 30    | 16   | Co-evolution loop gaps (7 items, 7 done) + existing P2 + 13 medium bugs from 2026-07-22 review + 4 latent collect->train bugs found closing the loop (1 fixed, 1 new HF-format gap logged) |
 | 🟢 P3    | 24    | 10   | Makefile, pre-commit, Docker, ADRs, ADR refresh complete; +10 hygiene items from 2026-07-22 review |
-| **Total** | **89** | **45** |
+| **Total** | **89** | **48** |
 
 > Update this table as you complete items. Recommended flow: P0 → P1 → P2 → P3.
 >
--- a/TODO.md
+++ b/TODO.md
@@ -209,7 +209,7 @@
 ### Medium-Priority Bugs Found in Whole-Repo Code Review (2026-07-22)
 
 - [x] **`bidirectional_manager.py` state fields are never mutated** _(done 2026-07-23)_ — fixed by `run_loop_cycle()` in item #6 above; `stats`, `evolution_history`, `is_running`, and `last_check_time` are now all mutated each cycle.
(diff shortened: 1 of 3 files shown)


SHA888:

1. ** POSIX-only**: Guarded with  — falls back to  on Windows so  won't silently fail.

2. **fd leak on early failure**: Set  after  takes ownership; the  handler now calls  when  (i.e. before  wrapped it).

3. ** narrowed to **: The  fallback now only catches the expected "file doesn't exist yet" case, letting real permission errors propagate.

All checks pass (ruff format, ruff check, 613 tests).

SHA888:
Addressed the review feedback (verified each point against the actual code before applying):

**1. TODO.md Total row (valid):** Updated Total Done from 45 → 48 (11+11+16+10=48). The pre-existing mismatch (47≠45 before this PR) was already wrong; the P2 bump just widened it.

**2. Symlink caveat (valid):** Added a docstring note to `_save_registry` documenting that `os.replace` operates on the path itself — callers pointing the registry at shared/mounted storage via a symlink should pass the resolved path or avoid symlinks. No code change (the old open-through-link behavior was also accidental, not intentional), but now it is at least documented.

**3. Cleanup-on-failure logging (valid):** The bare `except OSError: pass` in both cleanup blocks now logs: `os.close` failures at DEBUG level, `os.unlink` failures at WARNING level. Uses `%`-style lazy formatting per stdlib logger convention.

**4. Mode-detection broadened (valid):** `except FileNotFoundError` → `except OSError` so `PermissionError` or other stat failures fall back to `0o644` instead of aborting the save entirely.

**5. `__init__`/`_load_registry` calling `_save_registry` (verified, no change needed):** `_load_registry()` only reads or returns a default dict — it never calls `_save_registry()`. The first save happens from `register_version()`, so the test assumption that a brand-new file gets `0o644` is correct.

`ruff format --check .` ✅ | `ruff check evoseal/ tests/` ✅ | `pytest tests/` 613 passed ✅

PR #103, opened 2026-07-28
fix: remove redundant __init__ from CodeArchive model

SHA888:
## What

Removed the custom `__init__` from `CodeArchive` that manually re-implemented defaults already declared via Pydantic `Field(default_factory=...)` on the class attributes.

## Why

The `__init__` (previously lines 127-149) manually set defaults for `id`, `created_at`, `updated_at`, `version`, `tags`, `visibility`, `is_archived`, `dependencies`, and `metadata` — all of which are already declared via Pydantic field defaults on the class. This created **two sources of truth**: changing one set of defaults without the other silently produces inconsistent behavior (a drift hazard).

## What changed

- Removed the `def __init__(self, **data)` method from `CodeArchive`
- All defaults now flow through Pydantic's standard field-default machinery (which was already correct)

## Behavioral note

The one minor behavioral difference: the old `__init__` set `updated_at = created_at` when only `created_at` was supplied. With the Pydantic defaults, both `created_at` and `updated_at` independently default to `datetime.now(UTC)`. In practice this is more correct — the two timestamps reflect when the object was actually constructed, not a synthetic alias.

## Addresses

TODO.md P3 item: `models/code_archive.py:127-149` — `__init__` manually re-implements every default already provided by `Field(default_factory=...)`

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/models/test_code_archive.py -v` → 9 passed ✅
- `pytest tests/` (full suite) → 613 passed, 23 deselected ✅

## Summary by CodeRabbit

* **Refactor**
  * Simplified code archive object initialization by relying on built-in field defaults instead of custom construction logic.
  * Preserved the expected behavior for update timestamps: when an update time isn’t provided, it now reliably mirrors the creation time to prevent subtle timestamp drift.

* **Documentation**
  * Updated the maintenance checklist and tracking summary to reflect completion of one additional low-priority item.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -262,7 +262,7 @@
 - [ ] **`providers/local_models.py:103-119`** — `_query_installed_models` is `lru_cache`d indefinitely with no TTL; a newly pulled/removed Ollama model isn't picked up until something explicitly calls `clear_model_cache()`
 - [ ] **`model_fine_tuner.py:122-137`** — `_check_gpu_availability()` is defined but never called; despite the module docstring claiming "Requires a CUDA GPU," `initialize_model()` silently proceeds on CPU instead of failing fast
 - [ ] **`model_fine_tuner.py:220,240`** — `example['instruction']`/`example['output']` direct dict access with no `.get()`; a malformed training example raises `KeyError` surfaced only as a generic error string instead of a clear validation message
-- [ ] **`models/code_archive.py:127-149`** — `__init__` manually re-implements every default already provided by `Field(default_factory=...)`, a drift hazard (two sources of truth for the same defaults)
+- [x] **`models/code_archive.py:127-149`** — `__init__` manually re-implements every default already provided by `Field(default_factory=...)`, a drift hazard (two sources of truth for the same defaults)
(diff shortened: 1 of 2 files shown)


SHA888:
Addressed the reviewer's point about  drift.

**What was wrong:** The removed  set `updated_at = data["created_at"]` when not explicitly provided, guaranteeing byte-for-byte equality. With the field defaults, each has its own independent `default_factory=lambda: datetime.now(UTC)`, introducing microsecond drift.

**Fix:** Added `model_post_init` that sets `self.updated_at = self.created_at` on construction, restoring the original invariant. Verified:
- Fresh archive: `created_at == updated_at` ✓
- After `update()`: `updated_at > created_at` ✓  
- All 613 tests pass ✓

SHA888:
Addressed review feedback: `model_post_init` now guards the `updated_at = created_at` assignment with `if "updated_at" not in self.model_fields_set`, so explicit values survive deserialization (`model_validate` / `model_validate_json`). Verified: fresh construction still mirrors created_at; round-trip with distinct updated_at now preserves it. All 613 tests pass.

PR #104, opened 2026-07-28
fix: await provider health checks from running event loop

SHA888:
## What

`get_best_available_provider()` and `list_providers()` in `provider_manager.py` created a `loop.create_task(provider.health_check())` when inside a running event loop but **never awaited it**, then unconditionally set `is_healthy = True`. This silently treated unhealthy providers as healthy and leaked orphaned tasks.

## Why

The original code tried to bridge sync/async with `loop.create_task()` + a comment "For now, skip health check in running loop and assume healthy". This was a known gap left open. Providers that fail their health check (e.g. Ollama down) would still be selected and used, causing downstream errors.

## How

Added `_run_coro_sync()` — a small helper that:
- Uses `asyncio.run()` when no event loop is running (the common sync case)
- Delegates to a `ThreadPoolExecutor` when called from within a running loop, avoiding the `asyncio.run() cannot be called from a running event loop` error

Both `get_best_available_provider()` and `list_providers()` now use it and report the actual health-check result.

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/providers/test_provider_manager.py -q` → **10 passed** ✅
  - Covers: healthy/unhealthy selection, exception handling, list_providers health reporting, `_run_coro_sync` from both sync and async contexts

## Addresses

TODO.md item: "`provider_manager.py` treats unhealthy providers as healthy when called from a running event loop"

## Summary by CodeRabbit

* **Bug Fixes**
  * Provider health checks now run reliably in both synchronous and asynchronous contexts.
  * Best-provider selection no longer assumes health when checks are possible, and unhealthy providers are skipped.
  * Provider listings now report accurate health status and show health-check error details when checks fail.

* **Tests**
  * Added unit coverage for health-check execution (sync vs async), timeouts, exception handling, provider selection, and status reporting.

* **Documentation**
  * Updated task tracking to mark the provider health-check issue as completed and reflect progress.


SHA888:
Addressed all three review points. Here's the verification + disposition:

### 1. Docstring "not blocked" claim — ✅ Fixed
The reviewer was right. `pool.submit(asyncio.run, coro).result()` blocks the calling thread — `.result()` is a synchronous wait. The docstring was incorrect. Fixed to explain the real purpose (avoiding `asyncio.run() cannot be called from a running event loop` error) and explicitly note the caller *is* blocked.

### 2. Loop-bound resource risk (ThreadPoolExecutor + fresh event loop per call) — ℹ️ Caveat documented
Verified against actual code: `ollama_provider.health_check()` creates a fresh `aiohttp.ClientSession` per call via `async with` — no cached loop-bound resources. The reviewer correctly flagged this as theoretical ("Can't verify from the diff... flag as a caveat"). Added a docstring caveat noting the risk for future `health_check()` implementations that might cache resources.

### 3. ThreadPoolExecutor per provider in `list_providers()` — ✅ Fixed
The reviewer was right about the overhead. Added an optional `executor` parameter to `_run_coro_sync`. `list_providers()` now creates a single shared `ThreadPoolExecutor` (only when inside a running loop) and passes it to each `_run_coro_sync` call, with proper cleanup in a `finally` block. Added two new tests covering the shared executor behavior.

**Verification:** `ruff format --check .` ✅, `ruff check evoseal/ tests/` ✅, `pytest` — 636 passed ✅

PR #105, opened 2026-07-28
fix: use public execute_step API and handle running event loop in WorkflowAgent

SHA888:
## What

`WorkflowAgent.act()` called the private `WorkflowEngine._execute_step()` which uses `asyncio.run()` internally — raising `RuntimeError` when invoked from within an already-running event loop.

## How

- Added public `WorkflowEngine.execute_step()` that detects a running event loop and offloads to a `ThreadPoolExecutor` instead of re-entering `asyncio.run()`
- `WorkflowAgent.act()` now calls the public `execute_step()` method instead of the private `_execute_step()`
- Added `WorkflowAgent.act_async()` for callers already in an async context

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/agentic_system/ tests/test_workflow_engine.py -q` → **21 passed** ✅
  - New tests (10): sync context, running-loop context, error propagation, missing component, missing component field, agent state tracking, act_async, receive, get_status
  - Existing tests (11): all pass unchanged
- Full test suite: **633 passed** ✅

## Addresses

TODO.md item: "`agentic_workflow_agent.py:14` couples to a private API and can crash inside a running event loop"

## Summary by CodeRabbit

* **New Features**
  * Added a public single-step workflow execution API for both sync and async usage.
  * Workflow agents now support async actions (`act_async`) and track the latest result.
* **Bug Fixes**
  * Fixed workflow step execution when called from an already-running event loop.
  * Prevented deadlocks for nested and concurrent offloaded step execution.
  * Improved workflow cleanup to block new submissions and handle in-flight calls safely.
* **Tests**
  * Expanded unit coverage for sync/async contexts, error cases, lifecycle cleanup, and concurrency (including a cleanup race).
* **Documentation**
  * Updated EVOSEAL TODO tracking with the new fix details and refreshed completion counts.


SHA888:
Addressed 3 of 4 review points; the 4th was already valid. Details:

**Point 1 —  coupling to  (FIXED)**
Added a public  on  that wraps .  now calls the public method. New test  verifies the routing, and  covers the new public method directly.

**Point 2 — Per-call ThreadPoolExecutor creation (FIXED)**
Replaced the per-call `ThreadPoolExecutor(max_workers=1)` with a persistent `self._step_executor` on the engine, initialized in `__init__` and shut down in `cleanup()`. The executor is reused across calls, eliminating thread-spawn/teardown overhead in hot loops.

**Point 3 — Thread safety / event-loop concern (FIXED with lock)**
Added `self._step_lock = threading.Lock()` around the executor path. This serializes concurrent calls from multiple coroutines, preventing any future mutation of shared state from introducing race conditions. Verified that the current `_on_workflow_event` and `_publish_event` implementations only do logging (no loop-bound state), so the "attached to a different loop" failure mode doesn't apply today — but the lock is cheap insurance.

**Point 4 —  / pytest-asyncio (ALREADY VALID, no change needed)**
The reviewer flagged this as unverifiable from the diff, but: `pytest-asyncio` 1.3.0 is installed, `pytest.ini` sets `asyncio_mode = strict`, and `test_act_async` passes as-is. No fix needed.

All 635 tests pass, ruff clean.

SHA888:
## Review feedback addressed

Verified all four points against the committed code. Here's the verdict:

### 1. Deadlock on nested/composite steps (loop-offload path) — ✅ Fixed

The reviewer was wrong about *why* (the worker thread's ident *is* tracked — `_step_active_threads.add(wtid)` happens inside `_run()` before the step body executes), but right that the **committed** code had a deadlock: the offloaded path did `self._step_executor.submit(asyncio.run, self._execute_step_async(step)).result()` directly, with no `_run()` wrapper at all. A nested call from the step body would not find its thread in `_step_active_threads`, would fail re-entry detection, and would deadlock on `_step_lock` (held by the blocked caller) or re-submit to the occupied single-worker executor.

**Fix:** Wrapped the offloaded submission in a `_run()` closure that (a) adds the worker thread's ident to `_step_active_threads` before the step body runs, and (b) uses a `threading.Event` to signal readiness so the caller can wait after the worker thread is registered. Nested calls now detect re-entry and route to a temporary thread.

### 2. Inconsistent locking between `execute_step` and `execute_step_async` — ✅ Documented

`execute_step_async` does not (and should not) use `_step_lock` — asyncio is single-threaded, so two `await` calls cannot interleave mid-step. The gap is that callers using `asyncio.gather` or similar concurrency primitives need to self-serialize. Added a `.. note::` to the docstring explaining this. This is not a code bug but a documentation gap; the sync path's lock protects against cross-thread sync callers, not async-to-async races (which are the caller's responsibility).

### 3. Persistent background thread leak if `cleanup()` is skipped — ✅ Fixed

Added `__enter__`/`__exit__` context manager support so callers can use `with WorkflowEngine() as engine:` instead of manually calling `cleanup()`. Updated the `cleanup()` docstring to flag the mandatory-cleanup requirement and show the context manager usage.

### 4. Unsynchronized `_step_shutdown` flag (TOCTOU race) — ✅ Fixed

The committed code set `_step_shutdown = True` **after** `executor.shutdown(wait=False)`, outside any lock — a concurrent `execute_step` could slip past the check and hit the shut-down executor.

**Fix:** Moved the flag set **before** `executor.shutdown()`, under `_step_active_lock` (the same lock that `execute_step` uses to check it), preventing the TOCTOU race. Also changed `_step_lock` from `threading.Lock` to `threading.RLock` to avoid self-deadlock if re-entry occurs on the same thread as the lock holder.

### Verification

- `ruff format --check .` ✅
- `pytest` → **639 passed**, 24 deselected ✅
- 4 new tests: nested offloaded no-deadlock, concurrent serialization, cleanup prevents submissions, context manager cleanup

PR #106, opened 2026-07-29
feat(providers): add retry with exponential backoff to OllamaProvider

SHA888:
## What

`OllamaProvider.submit_prompt()` had no retry logic — a single transient network failure (timeout, connection reset, Ollama restart) would immediately propagate as an unhandled exception, crashing the calling evolution cycle.

## How

- Added `max_retries` (default 3) and `backoff_base` (default 1.0s) constructor parameters
- `submit_prompt()` now retries on transient failures: `TimeoutError`, `aiohttp.ClientError`, and HTTP 5xx responses
- Exponential backoff with jitter: `backoff_base * 2^(attempt-1) + random(0, 0.5)`
- Non-retryable errors (JSONDecodeError, HTTP 4xx, Ollama API errors like 'model not found') fail immediately
- Added `_is_retryable()` helper to classify transient vs permanent failures
- `get_model_info()` now includes retry configuration

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/providers/test_ollama_provider.py -q` → **16 passed** ✅
  - Tests: success path, success with system message, retry on timeout, retry on client error, retry on 500, exhaustion after max retries, no retry on bad JSON, no retry on API error, no retry on 400, _is_retryable classification, configuration, get_model_info

## Addresses

TODO.md item: "`providers/ollama_provider.py:98-136` — `submit_prompt` has no retry/backoff on transient network failures despite being the sole retry/timeout surface for local model calls"

## Summary by CodeRabbit

- **New Features**
  - Added configurable retries for transient prompt submissions, with exponential backoff and jitter.
  - Exposed retry configuration in model information.
- **Bug Fixes**
  - Improved resilience to timeouts, network errors, and server-side (5xx) failures.
  - Avoids retries for non-retryable outcomes such as invalid/undecodable responses and client-side (4xx) errors.
- **Tests**
  - Added unit coverage for request shaping, retry classification, retry/backoff timing, and exposed configuration.
- **Documentation**
  - Updated project tracking to mark the retry-support item as completed.


SHA888:
Addressed all three review points in bc6fb12:

**1. 5xx `continue` skipped backoff** — Removed the bare `continue`. 5xx responses now raise a typed `_OllamaServerError` exception, which flows through the unified `except` handler and hits the same exponential-backoff + jitter `asyncio.sleep` block as timeout/connection errors. Added `test_submit_prompt_500_retries_with_backoff` which mocks `asyncio.sleep` and asserts a positive delay is applied.

**2. `_is_retryable()` was dead code** — Wired it as the single retry-gate in `submit_prompt`'s `except` block. The hand-duplicated `isinstance(TimeoutError)` / `isinstance(aiohttp.ClientError)` checks are removed; `_is_retryable` now also recognizes `_OllamaServerError`. Added `test_is_retryable_server_error` for coverage.

**3. `assert last_exc is not None` for control flow** — Replaced with an explicit `if last_exc is None: raise RuntimeError(...)` so the invariant holds under `python -O`.

All 664 tests pass. Lint clean.

SHA888:

**1. Overall deadline across retries** — Added an overall deadline of `timeout * total_attempts` tracked via `asyncio.get_event_loop().time()`. The retry loop now aborts early if the remaining time is less than half of `self.timeout` (i.e., too short for a meaningful attempt). With defaults (timeout=120, max_retries=3), worst-case is still bounded at ~480s but is now explicitly tracked and documented rather than being an unbounded side effect.

**2. Backoff cap** — Added a `max_backoff` parameter (default 30s). The delay is now `min(backoff_base * 2^attempt, max_backoff) + jitter`, so with defaults the max delay per attempt is ~30.5s instead of growing unboundedly.

**3. Session reuse** — Hoisted the `aiohttp.ClientSession` out of the retry loop. A single session is created and reused across all attempts, eliminating the connection churn on retries.

Updated all retry tests to work with the single-session design (mock sessions now return sequenced responses instead of using `side_effect` to create a new session per attempt). All 22 ollama provider tests and 560 unit tests pass.

PR #107, opened 2026-07-29
fix(providers): add TTL-based expiry to installed-model cache

SHA888:
## What

`_query_installed_models` in `providers/local_models.py` used `functools.lru_cache` with no TTL — a newly pulled or removed Ollama model was never discovered until something explicitly called `clear_model_cache()`.  This is a P3 hygiene item from the whole-repo review (TODO.md).

## How

- Replaced `@lru_cache(maxsize=8)` with a time-based cache (`_model_cache` dict keyed by `(base_url, timeout)`) that auto-expires entries after `_CACHE_TTL_SECONDS` (default 120s)
- Cached failures too (stale-while-revalidate style) so a temporarily-down Ollama doesn't generate a flood of connection attempts
- Updated module and function docstrings to reflect TTL-based caching
- Added a test verifying TTL expiry triggers a fresh HTTP query

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/providers/test_local_models.py -q` → **34 passed** ✅
- Full test suite: **663 passed** ✅

## Addresses

TODO.md item: "`providers/local_models.py:103-119` — `_query_installed_models` is `lru_cache`d indefinitely with no TTL; a newly pulled/removed Ollama model isn't picked up until something explicitly calls `clear_model_cache()`"

## Summary by CodeRabbit

* **Bug Fixes**
  * Local model discovery now uses a time-limited cache to automatically detect newly added or removed models without manual cache clearing.
  * Transient lookup failures are cached briefly to avoid repeated request storms during outages.
  * The installed-model cache is bounded and supports concurrent refreshes so only one live lookup runs per endpoint.
* **Tests**
  * Added/expanded unit coverage for TTL expiry, failure-cache TTL behavior, cache-size bounds, cache-key normalization, stale-then-failing scenarios, and concurrent request coalescing (including malformed responses and error paths).

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -259,7 +259,7 @@
 - [x] **Stray `evoseal/utils/validator.py.bak`** _(done 2026-07-26)_ — a 1132-line backup committed to the repo tree (not gitignored), diverged from `validator.py`; risk of editing the wrong file
 - [x] **`evoseal/utils/testing/environment.py:180`** — `suffix: str = None` type-annotation mismatch (should be `Optional[str]`)
 - [ ] **`providers/ollama_provider.py:98-136`** — `submit_prompt` has no retry/backoff on transient network failures despite being the sole retry/timeout surface for local model calls
-- [ ] **`providers/local_models.py:103-119`** — `_query_installed_models` is `lru_cache`d indefinitely with no TTL; a newly pulled/removed Ollama model isn't picked up until something explicitly calls `clear_model_cache()`
+- [x] **`providers/local_models.py:103-119`** — `_query_installed_models` is now TTL-cached (120s default); a newly pulled/removed Ollama model is picked up automatically without needing `clear_model_cache()`
 - [ ] **`model_fine_tuner.py:122-137`** — `_check_gpu_availability()` is defined but never called; despite the module docstring claiming "Requires a CUDA GPU," `initialize_model()` silently proceeds on CPU instead of failing fast
 - [ ] **`model_fine_tuner.py:220,240`** — `example['instruction']`/`example['output']` direct dict access with no `.get()`; a malformed training example raises `KeyError` surfaced only as a generic error string instead of a clear validation message
 - [x] **`models/code_archive.py:127-149`** — `__init__` manually re-implements every default already provided by `Field(default_factory=...)`, a drift hazard (two sources of truth for the same defaults)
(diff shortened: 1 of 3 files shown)


SHA888:
Addressed both review points:

**1. Stale cache reuse on failure** — The failure branch was doing , which reused the expired *success* result when Ollama went down after a TTL expiry. This silently masked outages and contradicted the documented contract. Fixed: failures now unconditionally cache and return . Added test `test_success_then_expiry_then_failure_returns_empty` that covers the success → TTL expiry → failure path.

**2. Thread safety** — Replaced bare `dict` access with a `threading.Lock` (`_model_cache_lock`) protecting all reads and writes in `_query_installed_models`, `_cache_write`, and `clear_model_cache`. This restores the thread-safety guarantee that `functools.lru_cache` provided, since this module is reached from async code via thread-pool executors.

Verification:
- `ruff format --check .` ✅
- `pytest tests/unit/providers/test_local_models.py -q` → 39 passed ✅
- Full suite: 668 passed ✅

SHA888:

**1. TOCTOU race (duplicate concurrent HTTP requests)** — Fixed with per-key  coalescing. When multiple threads hit a cache miss for the same key, only one performs the blocking HTTP call; the others wait on the event and reuse the cached result. Different keys are not serialized against each other.

**2.  normalization** — The cache key now applies  to  before use, matching how the request URL is built.  and  now share one cache entry.

**3. LRU eviction** — Eviction now uses a separate  timestamp (refreshed on every cache hit) instead of the write timestamp, so frequently-accessed entries survive eviction regardless of age.

New tests:  (4 threads, barrier sync, verifies 1 HTTP call) and . All 41 tests pass, ruff clean.

SHA888:

**1. TOCTOU race (duplicate concurrent HTTP requests)** — Fixed with per-key threading.Event coalescing. When multiple threads hit a cache miss for the same key, only one performs the blocking HTTP call; the others wait on the event and reuse the cached result. Different keys are not serialized against each other.

**2. base_url normalization** — The cache key now applies rstrip("/") to base_url before use, matching how the request URL is built. "http://host:11434" and "http://host:11434/" now share one cache entry.

**3. LRU eviction** — Eviction now uses a separate last_access timestamp (refreshed on every cache hit) instead of the write timestamp, so frequently-accessed entries survive eviction regardless of age.

New tests: test_concurrent_calls_coalesce (4 threads, barrier sync, verifies 1 HTTP call) and test_trailing_slash_normalized_in_cache_key. All 41 tests pass, ruff clean.

SHA888:
Addressed both review points in ff94d88.

**Point 1 — Permanent deadlock from uncaught exceptions:**
Both issues confirmed against the actual code:
- `http.client.IncompleteRead` inherits from `HTTPException → Exception`, NOT `OSError` — a truncated read escapes the except clause entirely.
- If Ollama returns a JSON array/string instead of a dict, `payload.get("models", [])` raises `AttributeError` — also not caught.
- Either case skips both `event.set()` and `_in_flight.pop(key, None)`, and since `event.wait()` has no timeout, every thread coalescing on that key hangs forever — and the stale event poisons the key for all future calls too.

Fix: nested `try/except` + outer `try/finally` so cleanup always runs regardless of exception type. Also moved the result-building line into the inner try block.

**Point 2 — `_in_flight` unbounded growth:**
This was a consequence of point 1 (entries leaked due to missed cleanup). The `try/finally` fix ensures entries are always removed promptly after the HTTP call completes, so no size bound is needed — entries are inherently transient when cleanup is guaranteed.

**Verification:** ruff format ✓, ruff check ✓, 563 unit tests pass (5 skipped), 0 failures.

SHA888:

**1. Broadened except clause** — Added `http.client.HTTPException`, `AttributeError`, and `TypeError` to the caught exceptions in `_query_installed_models`. Malformed responses (non-dict `payload`, non-dict entries in `models` list) and HTTP-level errors like `IncompleteRead` are now handled the same as network failures: logged, cached as empty, and callers get `()` back instead of an uncaught crash.

**2. Identity-checked `_in_flight.pop`** — Changed from unconditional `pop(key, None)` to `if _in_flight.get(key) is event: pop(key, None)`. This prevents the scenario where a failed fetch (no `_cache_write`) leaves `_model_cache[key]` missing, a new thread registers a fresh in-flight marker for the same key, and the original fetcher's `finally` accidentally removes the *new* marker.

**3. Module-level TTL invariant assertion** — Added `assert _FAILURE_TTL_SECONDS < _CACHE_TTL_SECONDS` at module level (not just in tests) to guard the backdating formula `ts = now - _CACHE_TTL_SECONDS + _FAILURE_TTL_SECONDS` against future constant edits that could silently break it.

Tests added: `test_malformed_response_returns_empty`, `test_models_list_with_non_dict_entries_returns_empty`, `test_http_exception_returns_empty`, `test_failure_ttl_invariant_enforced`. Full suite: 674 passed.

PR #108, opened 2026-07-29
test: add regression tests for config validation edge cases

SHA888:
## What

Adds 9 new regression tests to `tests/unit/models/test_system_config.py` for `SystemConfig` validation edge cases:

- `test_from_yaml_file_not_found` — nonexistent path raises `FileNotFoundError`
- `test_from_yaml_comments_only_raises` — YAML with only comments parses as None, correctly rejected
- `test_validate_with_none_values_for_required_keys` — None values for required sections still validate (section key exists)
- `test_validate_extra_keys_accepted` — extra keys beyond REQUIRED_KEYS do not cause failure
- `test_get_through_non_dict_intermediate_returns_default` — dot-notation traversal through a non-dict intermediate returns default
- `test_get_with_numeric_key_in_path` — numeric-looking string keys in dot notation
- `test_get_none_value_vs_missing` — explicit None vs missing key distinction in get()
- `test_validate_multiple_missing_sections` — all missing sections reported in error
- `test_from_yaml_deeply_nested_config` — deeply nested YAML load and traversal

## Why

Addresses TODO.md P2 item: **Add regression test for config validation** — malformed YAML, missing required sections, type mismatches. These edge cases catch regressions in the from_yaml, validate, and get pipeline that the existing 7 tests do not cover.

## Verification

- ruff format --check: passed
- ruff check evoseal/ tests/: passed
- pytest tests/unit/models/test_system_config.py -v: 16/16 passed

## Summary by CodeRabbit

* **Tests**
  * Added regression coverage for configuration loading, validation, and nested value access.
  * Verified handling of missing files, comments-only YAML, deeply nested settings, optional values, extra keys, numeric keys, and missing paths.
  * Confirmed validation errors report all missing required sections.

* **Documentation**
  * Updated task tracking to reflect completion of configuration validation test coverage.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -199,7 +199,7 @@
 - [ ] **Increase unit test coverage for `core/` modules**
   - `controller.py`, `evaluator.py`, `selection.py`, `version_database.py`
   - Target: meaningful coverage on core logic paths, not just line count
-- [ ] **Add regression test for config validation**
+- [x] **Add regression test for config validation**
   - Malformed YAML, missing required sections, type mismatches
 - [ ] **Add test for checkpoint save/restore**
   - Interrupt mid-evolution, restore from checkpoint, verify state consistency
@@ -305,9 +305,9 @@
 |----------|-------|------|-------|
 | 🔴 P0    | 11    | 11   | Original 5 complete; all 6 critical bugs from 2026-07-22 whole-repo review fixed (PRs #74, #76-#79) |
 | 🟠 P1    | 24    | 14   | Original safety/integration items done; +12 high-priority bugs from 2026-07-22 review; signal-handler init fix; safety.yaml created |
-| 🟡 P2    | 30    | 19   | Co-evolution loop gaps (8 items, 8 done) + existing P2 + 13 medium bugs from 2026-07-22 review + 4 latent collect->train bugs found closing the loop (1 fixed, 1 new HF-format gap resolved); provider_manager health-check await fix; workflow-agent private-API/event-loop fix |
+| 🟡 P2    | 30    | 20   | Co-evolution loop gaps (8 items, 8 done) + existing P2 + 13 medium bugs from 2026-07-22 review + 4 latent collect->train bugs found closing the loop (1 fixed, 1 new HF-format gap resolved); provider_manager health-check await fix; workflow-agent private-API/event-loop fix |
 | 🟢 P3    | 24    | 15   | Makefile, pre-commit, Docker, ADRs, ADR refresh, CHANGELOG complete; +11 hygiene items from 2026-07-22 review; Ollama provider retry/backoff fix; local_models TTL cache |
-| **Total** | **89** | **59** | |
+| **Total** | **89** | **60** | |
 
 > Update this table as you complete items. Recommended flow: P0 → P1 → P2 → P3.
 >
--- a/tests/unit/models/test_system_config.py
+++ b/tests/unit/models/test_system_config.py
@@ -118,6 +118,22 @@ def test_validate_with_none_values_for_required_keys():
     assert sys_config.validate() is True
 
 
+def test_validate_with_type_mismatch_for_required_keys():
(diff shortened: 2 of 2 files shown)


SHA888:
Thanks for the review. Verified each point against the actual source:

**1. `yaml` import:** Already present — `import yaml` is on line 4 of the test file (pre-existing, not shown in the diff hunk). No fix needed.

**2. `None` values for required keys:** The actual `validate()` implementation checks key presence, not truthiness:
```python
missing = [k for k in self.REQUIRED_KEYS if k not in self.config]
```
`"dgm" in {"dgm": None}` is `True`, so `None` values correctly satisfy the required-section check. The test assertion is correct.

**3. Brittle exact substrings:** Acknowledged as a valid style concern. Both substrings match the actual implementation, and the new tests follow the exact same `match=` pattern used by the three pre-existing tests (`test_from_yaml_empty_file_raises`, `test_from_yaml_scalar_raises`, `test_from_yaml_list_raises`). Changing only the new tests would be inconsistent. If a broader cleanup is desired, that could be a separate follow-up.

SHA888:
Addressed review feedback:

**Point 1 (unverifiable assertions):** Verified all of them against the implementation — `"not found"`, `"did not produce a mapping"`, `get()`/`validate()` semantics all match `system_config.py` exactly. No changes needed.

**Point 2 (None values for required keys):** Confirmed this is the intended contract. `validate()` does `if k not in self.config` — a pure key-presence check, no type/truthiness gate. The test documents this correctly. Worth noting: if someone wants stricter validation later (rejecting `None` section values), that would be a new feature, not a bug fix.

**Point 3 (numeric key docstring):** Fixed. The old docstring claimed to test "numeric-looking string keys" but the test also contained `assert sys_config.get("port") == 8080` — a plain single-key lookup that doesn't exercise dot-notation at all. Removed that misleading assertion, expanded the docstring to explain what the test actually verifies (dict-key traversal with a numeric-looking string key, and why that distinction matters if list-index support is ever added). All 16 tests still pass.

SHA888:
Thanks for the thorough review. I verified each point against the actual `SystemConfig` implementation (`evoseal/models/system_config.py`):

### Point 1 — type-mismatch gap in TODO checkbox ✅ Fixed
Valid catch. The TODO item claims "type mismatches" but no test exercised it. Added `test_validate_with_type_mismatch_for_required_keys` — it sets `dgm` to a string, `openevolve` to a list, `seal` to an int, and `integration` to a bool, then confirms `validate()` still passes (it only checks key presence, not value type). This is the missing case between "present with `None`" and "missing entirely."

### Point 2 — brittle `match=` substrings ✅ Verified correct, no change
I checked the source directly:
- `from_yaml` raises `FileNotFoundError(f"YAML config file not found: {yaml_path}")` → `match="not found"` is accurate.
- `from_yaml` raises `ValueError(f"... did not produce a mapping (got {type(config_dict).__name__}).")` → `match="did not produce a mapping"` is accurate.

The pre-PR tests (`test_from_yaml_empty_file_raises`, `test_from_yaml_scalar_raises`, `test_from_yaml_list_raises`) already use the same `match="did not produce a mapping"` pattern, so this is established style in the file, not a new coupling introduced by this PR.

### Point 3 — behavioral assumptions in `get()` tests ✅ Verified correct, no change
The `get()` implementation is:
```python
if not isinstance(current, dict) or part not in current:
    return default
current = current[part]
return current
```
This confirms:
- **`test_get_none_value_vs_missing`**: Key present with `None` → `part in current` is True → returns `None`. Key missing → `part not in current` → returns default. Correct.
- **`test_get_through_non_dict_intermediate_returns_default`**: `not isinstance(current, dict)` catches the string → returns default. Correct.
- **`test_get_with_numeric_key_in_path`**: `part not in current` does string comparison on dict keys → `"0"` matches the string key `"0"`. Correct.

All three tests pass and their assertions match the actual behavior.

PR #109, opened 2026-07-29
fix: harden monitoring dashboard CORS and add optional auth

SHA888:
## What

Hardens `MonitoringDashboard` CORS and adds optional bearer-token authentication.

## Why

The monitoring dashboard had two security issues flagged in the whole-repo review (2026-07-22):

1. **Permissive CORS with credentials (CWE-942):** `setup_cors()` used wildcard `*` origin with `allow_credentials=True`, meaning any website could make credentialed requests to internal operational endpoints.
2. **No authentication:** All API endpoints returned internal data with zero auth — a problem if the dashboard is ever exposed beyond localhost.

## Changes

- **CORS:** Default origins now derive from `host:port` instead of `*`. Binding to `0.0.0.0` logs a security warning. When `*` is explicitly passed, `allow_credentials` is forced `False` per the CORS spec.
- **Auth:** New optional `auth_token` parameter. When set, all `/api/*` and `/ws` requests must present `Authorization: Bearer <token>`. WebSocket clients can pass `?token=` as a query param. The HTML dashboard page (`/`) is not gated.
- **14 new unit tests** covering: auth accept/reject for all endpoints, dashboard page not gated, CORS defaults, wildcard warning, `0.0.0.0` warning, and constructor parameter storage.

Addresses the P1 TODO item: "Monitoring dashboard has no authentication and permissive CORS-with-credentials".

## Summary by CodeRabbit

* **New Features**
  * Added optional bearer-token authentication for dashboard API routes and WebSocket clients (via `Authorization: Bearer <token>` and `Sec-WebSocket-Protocol` / `?token=`).
  * Added configurable cross-origin allowed origins and tightened defaults for publicly bound addresses.
* **Bug Fixes**
  * Improved CORS/auth compatibility: preflight (`OPTIONS`) is supported while keeping the dashboard landing page ungated.
  * Enforced safer token handling (constant-time comparison) and rejected invalid `auth_token` values.
* **Tests**
  * Expanded unit tests for constructor validation, CORS/auth behavior, WebSocket auth flows, and non-ASCII tokens.
* **Documentation**
  * Updated the security monitoring dashboard checklist to reflect completion.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -129,8 +129,9 @@
 - [x] **Model-safety validator can be defeated by an incidental safety word** _(done 2026-07-23, fix/safety-validator-bypass)_
   - `evoseal/fine_tuning/model_validator.py:432-454` — `_is_safe_response` returned `has_safety or not has_unsafe`. A response containing both an unsafe instruction and any safety-sounding word (e.g. `"Sorry, but here's how: rm -rf /"`) was classified safe regardless of the unsafe content. Fixed to `return not has_unsafe`; added 7 regression tests in `test_model_validator.py`
 - [x] **Path traversal in git-file read/write helpers** _(done 2026-07-26)_ — `get_file_content` and `write_file_content` now resolve the constructed path and verify it stays within `repo_path` via `is_relative_to()`, raising `GitError` on traversal. 3 new test functions cover `..` traversal, absolute paths, nested traversal, and legitimate subdirectory ops.
-- [ ] **Monitoring dashboard has no authentication and permissive CORS-with-credentials**
-  - `evoseal/services/monitoring_dashboard.py:75-91` (`setup_cors`) — every route gets `aiohttp_cors.ResourceOptions(allow_credentials=True, allow_headers="*", allow_methods="*", expose_headers="*")` on a wildcard origin (CWE-942). No auth on any HTTP or WebSocket endpoint (`/api/status`, `/api/metrics`, `/api/report`, `/ws`), which return internal operational data (data paths, config, error strings). Defaults to `localhost` (limits blast radius today) but nothing prevents/warns against a `0.0.0.0` deploy, at which point this is unauthenticated remote information disclosure
+- [x] **Monitoring dashboard has no authentication and permissive CORS-with-credentials** _(done 2026-07-29)_
+  - `evoseal/services/monitoring_dashboard.py:75-91` (`setup_cors`) — every route got `aiohttp_cors.ResourceOptions(allow_credentials=True, allow_headers="*", allow_methods="*", expose_headers="*")` on a wildcard origin (CWE-942). No auth on any HTTP or WebSocket endpoint (`/api/status`, `/api/metrics`, `/api/report`, `/ws`), which return internal operational data (data paths, config, error strings). Defaults to `localhost` (limits blast radius today) but nothing prevented/warned against a `0.0.0.0` deploy, at which point this was unauthenticated remote information disclosure
+  - Fixed: CORS origins now default to the dashboard's own host:port (not `*`); binding to `0.0.0.0` logs a security warning; added optional `auth_token` parameter — when set, all `/api/*` and `/ws` requests must present `Authorization: Bearer <token>` (WebSocket clients can pass `?token=`); the HTML dashboard page is not gated. Wildcard `*` origin explicitly disables `allow_credentials` per CORS spec. 14 new unit tests cover auth accept/reject, CORS defaults, wildcard warning, and constructor params
 
 ### CI/CD & Release Pipeline Issues Found in Whole-Repo Code Review (2026-07-22)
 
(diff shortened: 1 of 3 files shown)


SHA888:
Addressed the review feedback. Here is the point-by-point verdict and changes:

**1. Constant-time token comparison** ✅ Fixed
- `auth_header == f"Bearer {self.auth_token}"` and `token != self.auth_token` replaced with `hmac.compare_digest` in both places. Added `import hmac`.

**2. OPTIONS preflight not exempted** ✅ Fixed
- Added `if request.method == "OPTIONS": return await handler(request)` early in `_auth_middleware`, before the auth check. CORS preflight requests now pass through to aiohttp_cors.

**3. `allowed_origins=["*"]` + `auth_token` combo not enforced** ✅ Fixed
- Added a runtime `logger.warning` in `__init__` when both are set, matching the existing warning pattern. The docstring guard rail is now surfaced at runtime.

**4. `logger` import/definition** ❌ Not a bug
- `logger = logging.getLogger(__name__)` is already defined at module scope (line 19). No change needed.

**Verification:** `ruff format --check` ✅, `ruff check` ✅, `pytest tests/unit/services/test_monitoring_dashboard.py` — 17 passed (including 3 new tests for OPTIONS passthrough, `::` bind warning, and wildcard+auth combo warning).

SHA888:
Addressed review feedback (commit 0c20042):

**Changed:**
1. ** +  now raises ** instead of only logging a warning. Operators cannot silently end up with credentialed cross-origin access wide open. Updated test accordingly.

2. **Documented  default limitation** in the  docstring — notes that explicit origins are needed when behind TLS termination or accessed via a different hostname than the bind host.

**Assessed but not changed:**
3. **WebSocket `?token=` query param** — browsers cannot set custom headers on WebSocket connections (the `WebSocket` API only accepts a URL + optional protocols). The query param approach is the standard pattern. WS subprotocol injection is possible but adds complexity and still ends up in the URL. No change needed.

4. **Dashboard page leaking `auth_token` into embedded JS** — verified that `_generate_dashboard_html()` never interpolates the token into HTML or JavaScript. The JS client opens a WebSocket to `/ws` with no token and does `fetch('/api/metrics')` with no auth header. The token is not leaked. (Note: this means the JS fallback fetch will get 401'd when auth is enabled — a functional limitation, not a security issue.)

SHA888:
Addressed the review feedback. Here's my assessment of each point and what changed:

### 1. WebSocket token in query string — ✅ Fixed
The `_auth_middleware` now prefers `Sec-WebSocket-Protocol` header over the `?token=` query param. The header value is parsed as a comma-separated list of subprotocols; entries starting with `Bearer ` are treated as the auth token. The `?token=` fallback is still accepted (for clients that can't set WS headers) but is documented as less secure since it leaks into access/proxy logs.

The embedded JavaScript client now reads the token from the page URL query param (`/?token=xxx`) and passes it via the `Sec-WebSocket-Protocol` header, so the token never appears in server WebSocket access logs. The HTTP API polling also sends the token via `Authorization` header.

### 2. Wildcard bind only logs warning — ⏭️ Intentionally unchanged
Raising an error on `0.0.0.0`/`::` would break legitimate behind-firewall deployments. The warning is clear and actionable ("Ensure auth_token is set or restrict access via firewall"). The existing test `test_bind_all_warns` confirms this is the intended behavior.

### 3. `hmac.compare_digest` non-ASCII TypeError — ✅ Fixed
Added `_constant_compare(a, b)` static method that encodes both operands to UTF-8 bytes before calling `hmac.compare_digest`, then catches `UnicodeEncodeError`/`AttributeError` and returns `False` instead of letting a `TypeError` bubble into a 500. Used for all auth comparisons (Authorization header, WS protocol token, query param token). New tests: `test_non_ascii_token_does_not_500` (round-trip with `über-sécret` token) and `test_constant_compare_non_ascii`.

### 4. CORS bare `http://localhost` without port — ✅ Fixed
Removed the bare `http://localhost` entry from the default origins for `0.0.0.0`/`::` binds. Only `http://localhost:<port>` is kept. Updated test assertion confirms the bare entry is excluded.

### 5. CORS per-origin `defaults` behavior — ✅ Verified, no change needed
Confirmed `aiohttp_cors.setup(defaults={origin: opts for origin in origins})` works identically to `defaults={"*": opts}` for `cors.add(route)` calls — each route gets CORS handling for each origin in the mapping. The OPTIONS preflight is handled by aiohttp_cors middleware regardless of whether defaults uses `*` or specific origins. The library's own docs show per-origin defaults as the standard pattern.

SHA888:
Addressed all four review points:

**1. IPv6 default-origin bracketing** — Added  helper using . Now  produces  instead of the invalid . Tests: , , .

**2. Per-origin ** — Replaced the single shared  with per-origin computation in . Wildcard  gets ; explicit origins keep . Test: .

**3. WebSocket subprotocol echo** —  now reads the  header, extracts the  subprotocol, and passes it to  so browsers complete the handshake per RFC 6455 §4.2.2. Tests: , .

**4. Token charset validation** — Added  regex check (RFC 7230 §3.2.6 tchar) in . Tokens containing , `=`, whitespace, or non-ASCII characters now raise  at construction time, preventing silent fallback to the leaky query-param path. Existing  updated to  (non-ASCII is also invalid tchar). New tests: , , .

All 32 tests pass. Lint clean.

SHA888:
Addressed all four review points:

**1. IPv6 default-origin bracketing** — Added `_bracket_ipv6()` helper using `ipaddress.ip_address()`. Now `host="::1"` produces `http://[::1]:8081` instead of the invalid `http://::1:8081`. Tests: `test_ipv6_loopback_origin_is_bracketed`, `test_ipv6_custom_origin_is_bracketed`, `test_ipv4_origin_not_double_bracketed`.

**2. Per-origin `allow_credentials`** — Replaced the single shared `default_opts` with per-origin computation in `setup_cors()`. Wildcard `*` gets `allow_credentials=False`; explicit origins keep `allow_credentials=True`. Test: `test_mixed_wildcard_and_explicit_origins_credentials_per_origin`.

**3. WebSocket subprotocol echo** — `websocket_handler` now reads the `Sec-WebSocket-Protocol` header, extracts the `bearer.<token>` subprotocol, and passes it to `WebSocketResponse(protocols=...)` so browsers complete the handshake per RFC 6455 S4.2.2. Tests: `test_ws_echoes_bearer_subprotocol`, `test_ws_no_subprotocol_when_no_auth`.

**4. Token charset validation** — Added `_is_http_token()` regex check (RFC 7230 S3.2.6 tchar) in `__init__`. Tokens containing `/`, `=`, whitespace, or non-ASCII characters now raise `ValueError` at construction time, preventing silent fallback to the leaky query-param path. Existing `test_non_ascii_token_does_not_500` updated to `test_non_ascii_token_rejected_at_init` (non-ASCII is also invalid tchar). New tests: `test_invalid_auth_token_with_slash_raises`, `test_invalid_auth_token_with_space_raises`, `test_valid_urlsafe_token_accepted`.

All 32 tests pass. Lint clean.

PR #110, opened 2026-07-30
fix: call _check_gpu_availability() in initialize_model to fail fast without GPU

SHA888:
## What

`initialize_model()` in `model_fine_tuner.py` now calls the existing `_check_gpu_availability()` method before attempting to load a model. If no CUDA GPU is available, it returns `False` with a clear error message directing users to the prompt-level path (`evoseal.prompt_evolution`).

## Why

The method `_check_gpu_availability()` was defined (lines 122-137) but never called. The module docstring explicitly says "Requires a CUDA GPU" and directs CPU-only hosts to use the prompt-level path, but `initialize_model()` silently proceeded on CPU — potentially very slow or producing unexpected results.

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/fine_tuning/ -q` → 69 passed ✅

Closes TODO.md item: `model_fine_tuner.py:122-137`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fine-tuning now verifies NVIDIA CUDA GPU availability before model initialization and fails fast with a clear error when none is available.
  * GPU checks are now explicitly limited to NVIDIA CUDA; non-CUDA paths are directed to the CPU/prompt-based workflow.
* **Documentation**
  * Updated the task tracking checklist and completed item counts to reflect the new GPU-availability handling.
* **Tests**
  * Added unit test coverage to ensure initialization is skipped and model/tokenizer loading does not occur when GPU availability is unavailable.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -260,7 +260,7 @@
 - [x] **`evoseal/utils/testing/environment.py:180`** — `suffix: str = None` type-annotation mismatch (should be `Optional[str]`)
 - [x] **`providers/ollama_provider.py:98-136`** — `submit_prompt` has no retry/backoff on transient network failures despite being the sole retry/timeout surface for local model calls
 - [x] **`providers/local_models.py:103-119`** — `_query_installed_models` is now TTL-cached (120s default); a newly pulled/removed Ollama model is picked up automatically without needing `clear_model_cache()`
-- [ ] **`model_fine_tuner.py:122-137`** — `_check_gpu_availability()` is defined but never called; despite the module docstring claiming "Requires a CUDA GPU," `initialize_model()` silently proceeds on CPU instead of failing fast
+- [x] **`model_fine_tuner.py:122-137`** — `_check_gpu_availability()` was defined but never called; `initialize_model()` now calls it and returns `False` with a clear error when no CUDA GPU is available
 - [ ] **`model_fine_tuner.py:220,240`** — `example['instruction']`/`example['output']` direct dict access with no `.get()`; a malformed training example raises `KeyError` surfaced only as a generic error string instead of a clear validation message
 - [x] **`models/code_archive.py:127-149`** — `__init__` manually re-implements every default already provided by `Field(default_factory=...)`, a drift hazard (two sources of truth for the same defaults)
 - [x] **`evoseal/agents/agentic_system_example.py:7`** — `from evoseal.agentic_system import ...` is the wrong module path (actual: `evoseal.agents.agentic_system`); the example fails immediately with `ModuleNotFoundError`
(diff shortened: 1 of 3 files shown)


SHA888:
Reviewed both points against the actual implementation — neither applies:

**Point 1 (exception propagation):** `_check_gpu_availability()` (line 122) already has its own `try/except ImportError` block wrapping the `import torch` and CUDA calls. On a CPU-only host, `import torch` succeeds (torch installs fine on CPU) and `torch.cuda.is_available()` returns `False` — no exception propagates. The `except ImportError` catches the case where torch is not installed at all. The described scenario (unhandled traceback on CPU-only) does not occur with this implementation.

**Point 2 (sync vs async):** The method is `def _check_gpu_availability(self) -> bool:` — a plain synchronous method, not `async def`. The call `if not self._check_gpu_availability():` evaluates the bool return directly. No `await` needed; the coroutine-object trap does not apply here.

No changes required.

SHA888:
## Review feedback addressed

**Point 1 — CUDA-only check (docstring + error message):** ✅ Fixed
- `_check_gpu_availability()` docstring now explicitly states CUDA-only and explains why (ROCm/MPS not supported — the fine-tuning path relies on CUDA-specific features like `torch.cuda`, `device_map="auto"`, fp16).
- Error message updated from "requires a GPU" to "requires a CUDA GPU (ROCm and MPS are not currently supported)".

**Point 2 — Exception handling in `_check_gpu_availability()`:** ✅ Fixed
- Broadened `except ImportError` → `except Exception`. A broken CUDA install (driver mismatch, runtime probe failure) would previously escape as an unhandled error since the call sits *outside* `initialize_model`'s `try` block. Now those failures are caught and return `False` like any other "no GPU" case.

**Point 3 — Test coverage for CPU-only early exit:** ✅ Added
- Two new regression tests in `TestInitializeModelGpuGuard`:
  - `test_returns_false_when_gpu_unavailable` — asserts `initialize_model()` returns `False` and `is_initialized` stays `False`.
  - `test_does_not_load_model_when_gpu_unavailable` — asserts `AutoTokenizer.from_pretrained()` and `AutoModelForCausalLM.from_pretrained()` are never called.

**Point 4 — TODO.md totals:** ✅ Verified, no issue (P3 15→16, Total 60→61, math checks out).

All ruff checks and tests pass locally.

SHA888:
Addressed review feedback (2 of 3 points — the third was already handled):

1. **`logger.debug` → `logger.warning`** in `_check_gpu_availability` exception handler. A broken torch install or driver permission error is now visible at default log levels instead of being silently swallowed behind the generic "No CUDA GPU available" message.

2. **Unguarded `import torch` → `pytest.importorskip("torch")`** in `test_gpu_check_returns_false_on_runtime_error`. The test now skips cleanly when torch isn't installed instead of erroring out.

3. **Existing tests calling `initialize_model()`** — verified: the two existing tests in `test_model_fine_tuner.py` both already mock `_check_gpu_availability` to return `False`, and no other test file calls `initialize_model()`. No change needed.

PR #111, opened 2026-07-30
fix: validate training example keys and use .get() in model_fine_tuner

SHA888:
## What

Malformed training examples (missing `instruction` or `output` keys) raised a raw `KeyError` with no useful context. Added upfront validation that skips invalid examples with clear log warnings, and switched all direct dict access to `.get()` with empty-string defaults.

## Why

`model_fine_tuner.py:220,240` (and line 294 in `_tokenize_if_needed`) used `example[instruction]` and `example[output]` — direct dict access that raises `KeyError` for any malformed training example, surfaced only as a generic error string instead of a clear validation message.

## Changes

- Added validation loop after loading examples: checks each is a dict with required keys (`instruction`, `output`), skips invalid entries with `logger.warning`, fails fast if zero valid examples remain
- Changed all 3 format-string sites from `example[key]` to `example.get(key, )`

## Verification

- `ruff format --check .` ✅ (399 files formatted)
- `pytest tests/unit/fine_tuning/ -q` → 69 passed ✅

Closes TODO.md item: `model_fine_tuner.py:220,240`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved fine-tuning training-data validation by checking required fields and optional input values.
  * Invalid examples are now filtered consistently, with clear errors when no usable training data remains.

* **Tests**
  * Expanded coverage for valid, invalid, mixed, and empty training datasets.

* **Documentation**
  * Updated task tracking to reflect completion of the training-data validation improvement.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -261,7 +261,7 @@
 - [x] **`providers/ollama_provider.py:98-136`** — `submit_prompt` has no retry/backoff on transient network failures despite being the sole retry/timeout surface for local model calls
 - [x] **`providers/local_models.py:103-119`** — `_query_installed_models` is now TTL-cached (120s default); a newly pulled/removed Ollama model is picked up automatically without needing `clear_model_cache()`
 - [ ] **`model_fine_tuner.py:122-137`** — `_check_gpu_availability()` is defined but never called; despite the module docstring claiming "Requires a CUDA GPU," `initialize_model()` silently proceeds on CPU instead of failing fast
-- [ ] **`model_fine_tuner.py:220,240`** — `example['instruction']`/`example['output']` direct dict access with no `.get()`; a malformed training example raises `KeyError` surfaced only as a generic error string instead of a clear validation message
+- [x] **`model_fine_tuner.py:220,240`** — `example['instruction']`/`example['output']` direct dict access with no `.get()`; a malformed training example raises `KeyError` surfaced only as a generic error string instead of a clear validation message
 - [x] **`models/code_archive.py:127-149`** — `__init__` manually re-implements every default already provided by `Field(default_factory=...)`, a drift hazard (two sources of truth for the same defaults)
 - [x] **`evoseal/agents/agentic_system_example.py:7`** — `from evoseal.agentic_system import ...` is the wrong module path (actual: `evoseal.agents.agentic_system`); the example fails immediately with `ModuleNotFoundError`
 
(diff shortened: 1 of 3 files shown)


SHA888:
Addressed the review feedback on `prepare_training_data` validation.

**Reviewer's point:** Validation only checked that `instruction`/`output` keys were *present*, not that their values were non-null/non-empty strings. `example.get('instruction', '')` returns `None` (not `''`) when the key exists with value `None`, producing literal `"None"` text in training data.

**Fix:** Added a value-type check after the key-presence check — each required key must be a non-empty string (`isinstance str` and `strip` non-empty). Examples with `None`, empty-string, or non-string values are now skipped with a clear warning.

**Verification:**
- `ruff format --check .` ✅
- `pytest` → 707 passed ✅
- Pre-commit hooks (detect-secrets) ✅

SHA888:
Addressed review feedback with f208a57:

**Point 1 — `_tokenize_if_needed` missing validation:** ✅ Fixed. Extracted `_validate_training_examples()` as a module-level helper and call it from both `prepare_training_data()` and `_tokenize_if_needed()`. The reviewer's caveat was correct — `_tokenize_if_needed` receives datasets from `TrainingDataBuilder` (via `fine_tune_model` → `load_from_disk`), which never passed through the validation in `prepare_training_data`. Now malformed examples are skipped with a warning in both paths.

**Point 2 — Redundant `.get()` post-validation:** ✅ Fixed. Switched to direct `example['instruction']` / `example['output']` indexing in the fallback and formatted-examples branches of `prepare_training_data()`. Validation already guarantees these keys are present non-empty strings, so `.get()` defaults were masking potential future regressions. Kept `.get()` for the optional `input` field (may be absent).

**Error message wording:** ✅ Fixed. Changed from 'all missing required keys' to 'No valid training examples remain after validation' — covers non-dict entries and non-string/empty values too.

**Verification:**
- `ruff format --check .` ✅
- `pytest` → 600 passed, 5 skipped ✅

SHA888:
## Addressed review feedback

### Point 1: Silent return on all-invalid examples ✅ Fixed
**Reviewer was correct.** `_tokenize_if_needed` returned the original untokenized dataset when all examples failed validation — the exact failure mode this PR was meant to eliminate. Changed to raise `ValueError` instead, which the caller (`fine_tune_model`) already catches and converts to `{"success": False, "error": ...}`.

### Point 2: Stub `Dataset` + conditional import — Already guarded, no runtime bug
**Reviewer was cautious** ("Can't confirm from this diff alone") and their concern was reasonable, but the call site is safe. `_tokenize_if_needed` is only called from `fine_tune_model` (line 436), which returns early with a fallback path when `TRANSFORMERS_AVAILABLE` is False (lines 418–433). The stub `Dataset` never reaches this function, and `from datasets import Dataset` is always available at that point. **No change needed** — the conditional import was also removed as part of the fix below.

### Point 3: Memory duplication via `to_list()` ✅ Fixed
**Reviewer was correct.** Replaced `dataset.to_list()` → `_validate_training_examples()` → `_RealDataset.from_list(valid)` with `dataset.filter()`, which avoids materializing the entire dataset into a Python list. Also removed the now-unnecessary `from datasets import Dataset as _RealDataset` import.

### Verification
- `ruff format --check .` ✅
- `pytest tests/unit/fine_tuning/ -q` → 5 passed ✅ (added 2 new tests for empty-validation and mixed-valid/invalid cases)
- Full pre-commit suite (709 tests) ✅

SHA888:
Addressed the review feedback. Here's the verdict on each point:

**1. Duplicated validation logic** ✅ Fixed

Extracted `_is_valid_example()` as a module-level predicate. Both `_validate_training_examples()` and `_tokenize_if_needed` now use it — `_tokenize_if_needed` calls it directly via `dataset.filter(_is_valid_example)`, and `_validate_training_examples` delegates to it for the actual check while preserving per-example log messages. The two paths can no longer drift apart.

**2. `.filter()` on stub `Dataset`** ❌ Not an issue — verified safe

`_tokenize_if_needed` is only called from `fine_tune_model()` (line 445), which is guarded by `if not TRANSFORMERS_AVAILABLE or not self.is_initialized:` that returns early with a fallback. So the stub `Dataset` (with no `.filter()`) never reaches this path. No change needed.

**3. `input` field unvalidated** ✅ Fixed

Added type check for `input` to `_is_valid_example()`: if `input` is present but not a string (e.g. `None`, list, number), the example is rejected. This is now enforced on both the `prepare_training_data` path (via `_validate_training_examples`) and the `_tokenize_if_needed` path (via the shared predicate). Added parametrized tests covering `None`, list, and int input values.

**Verification:** `ruff format --check .` ✅ | `ruff check evoseal/ tests/` ✅ | `pytest tests/unit/fine_tuning/ -q` → 22 passed ✅

PR #114, opened 2026-07-31
fix: default trust_remote_code to False in model_fine_tuner

SHA888:
## What

Changed both `AutoTokenizer.from_pretrained` and `AutoModelForCausalLM.from_pretrained` calls from `trust_remote_code=True` to `trust_remote_code=False` (the safe default). The `# nosec B615` suppression comments remain because B615 flags missing `revision=`, not `trust_remote_code` — revision pinning is left to the caller.

## Why

`trust_remote_code=True` allows a HuggingFace repo to execute arbitrary Python code on model load. Combined with the `_resolve_hf_base_model()` fallback that uses `model_name` verbatim as an HF repo id for unknown families, a bad config value or env var (`EVOSEAL_CODER_MODEL`) could execute arbitrary remote code locally.

The known HF base models (DeepSeek, Qwen, CodeLlama, Mistral) are standard architectures that do not require remote code execution — `trust_remote_code=False` is the correct default.

## Changes

- `model_fine_tuner.py`: `trust_remote_code=True` → `False` in both loading calls
- `test_model_fine_tuner.py`: Added 2 regression tests — mock-based verification that both `from_pretrained` calls receive `trust_remote_code=False`, plus behavior check when TRANSFORMERS_AVAILABLE is False

## Verification

- `ruff format --check .` ✅ (400 files formatted)
- `ruff check evoseal/ tests/` ✅ All checks passed
- `pytest tests/unit/fine_tuning/test_model_fine_tuner.py -q` → 5 passed ✅

Closes TODO.md item: `model_fine_tuner.py:160-178`

## Summary by CodeRabbit

- **Security**
  - Improved model and tokenizer loading by disabling execution of untrusted remote code by default.

- **Bug Fixes**
  - Reduced security risk during fine-tuning setup.
  - Preserved expected behavior when the Transformers package is unavailable.

- **Tests**
  - Added coverage confirming remote code execution remains disabled for both loading operations.
  - Updated tokenization and security-related test documentation.


SHA888:
Addressed review feedback:

**1. TODO.md changelog mismatch** — Fixed. The entry incorrectly said "removed the `# nosec B615` suppression comments." They were relocated, not removed. B615 flags missing `revision=`, not `trust_remote_code` — the suppressions are still needed. Corrected the TODO.md entry and the PR body.

**2. `# nosec B615` moved to closing `)` line** — Fixed. Moved the comment back to the opening `AutoTokenizer.from_pretrained(` line where bandit reports the issue. The model call was already on one line so its placement was fine.

**3. Hardcoding `trust_remote_code=False` removes override path** — Acknowledged but not changed in this PR. The original TODO item was specifically to close the RCE vector. The known model families (DeepSeek, Qwen, CodeLlama, Mistral) are standard architectures. Adding a config knob for `trust_remote_code` would be scope creep — if someone needs it for a specific model, that should be a separate feature with proper allowlist semantics.

**4. Fragile source-inspection test** — Fixed. Replaced `inspect.getsource` + substring matching with a mock-based test that patches `from_pretrained` on both classes and asserts `trust_remote_code=False` in each call's kwargs — consistent with the existing `test_passes_max_length_through_to_tokenizer` style. Also fixed `test_initialize_returns_false_without_transformers` to actually patch `TRANSFORMERS_AVAILABLE=False` via `monkeypatch` instead of relying on the environment accidentally lacking `peft`.

Verification: `ruff format` ✅ `ruff check` ✅ `pytest` 5/5 ✅

SHA888:
## Review feedback response

### Point 1: Hardcoded `trust_remote_code=False` — consider allowlist/opt-in

**Pushing back on this one.** All5 currently-mapped base models use standard transformers architectures and don't need `trust_remote_code=True`:

| Model | Architecture | Needs custom code? |
|-------|-------------|-------------------|
| `deepseek-ai/deepseek-coder-6.7b-instruct` | LlamaForCausalLM | No |
| `Qwen/Qwen2.5-Coder-7B-Instruct` | Qwen2ForCausalLM | No |
| `codellama/CodeLlama-7b-Instruct-hf` | LlamaForCausalLM | No |
| `mistralai/Mistral-7B-Instruct-v0.2` | MistralForCausalLM | No |

(Confirmed via HF API — all tagged `library_name: transformers` with standard `model_type`.)

Adding an opt-in `trust_remote_code` parameter to `__init__` would reintroduce the exact RCE vector this PR closes — any caller (or misconfigured env var) could flip it back to `True`. The `_resolve_hf_base_model()` unknown-family fallback (L116-120) that uses `model_name` verbatim as an HF repo id is the attack surface; `trust_remote_code=False` is the fix.

If a future model genuinely needs custom code, the right path is: add it to `HF_BASE_MODEL_BY_FAMILY` with a pinned revision, or the operator sets `base_model_path` to a local path they've already vetted. That's a separate feature, not a regression from this change.

### Point 2: Tests don't exercise the failure path

**Valid — addressed.** Added `test_initialize_returns_false_when_model_needs_trust_remote_code` which:
- Mocks `AutoTokenizer.from_pretrained` to raise `OSError` (what transformers actually raises when the auto-map can't resolve architecture without custom code)
- Asserts `initialize_model()` returns `False` and `is_initialized` stays `False`
- Asserts model load was never attempted (tokenizer fails first)
- Asserts the error is logged at `ERROR` level with `"Error initializing model"` for operator diagnosis

### Point 3: How `from_pretrained` exceptions are surfaced

**Already handled.** `initialize_model()` wraps the entire body in `try/except Exception` (L186-188) which catches the error, logs it via `logger.error(f"Error initializing model: {e}")`, and returns `False`. The new failure-path test explicitly asserts this log message is emitted. The full method body is visible in the diff — there's no hidden code path that silently swallows the exception.

PR #115, opened 2026-07-31
test: increase unit test coverage for core/ modules

SHA888:
## What

Add 65 new tests across 4 test files covering `controller.py`, `evaluator.py`, `selection.py`, and `version_database.py` — the modules called out in TODO.md P2 "Increase unit test coverage for `core/` modules".

## New test files

- **`test_controller.py`** (12 tests): initialize, run_generation orchestration, select_candidates sorting/truncation, get_state, cli_interface dispatch
- **`test_version_database_extended.py`** (26 tests): experiment tracking (add/get by experiment), get_variant_metadata, get_best_variants with filtering/limits, get_variant_statistics (empty, same-score, per-experiment), export/import (JSON string, file, duplicates, lineage preservation, experiment tracking), edge cases
- **`test_selection_extended.py`** (14 tests): custom fitness_key, empty population (raises IndexError), single individual, elitism > population size, elitism=0, num_selected=0 and >population, custom strategies, negative/missing scores
- **`test_evaluator_extended.py`** (13 tests): empty results, missing/partial metrics, unknown strategy fallback, all-metrics-perfect/zero, feedback content (flags low coverage/quality, clean when perfect), custom default_weights, strategy overwrite, multiple results ordering

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/evoseal/ -v` → 81 passed ✅ (65 new + 16 existing)

Closes TODO.md item: "Increase unit test coverage for `core/` modules"

## Summary by CodeRabbit

* **Tests**
  * Expanded unit test coverage for controller workflows, evaluation strategies, candidate selection, and version database operations.
  * Added coverage for edge cases including empty data, missing metrics, boundary scores, duplicate records, filtering, invalid commands, and custom strategies.
  * Verified experiment tracking, metadata, statistics, lineage preservation, and JSON/file import and export.
  * Increased tracked test coverage completion from 64 to 65 overall.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -197,7 +197,7 @@
 
 ### Testing Coverage
 
-- [ ] **Increase unit test coverage for `core/` modules**
+- [x] **Increase unit test coverage for `core/` modules**
   - `controller.py`, `evaluator.py`, `selection.py`, `version_database.py`
   - Target: meaningful coverage on core logic paths, not just line count
 - [x] **Add regression test for config validation**
@@ -307,9 +307,9 @@
 |----------|-------|------|-------|
 | 🔴 P0    | 11    | 11   | Original 5 complete; all 6 critical bugs from 2026-07-22 whole-repo review fixed (PRs #74, #76-#79) |
 | 🟠 P1    | 24    | 15   | Original safety/integration items done; +12 high-priority bugs from 2026-07-22 review; signal-handler init fix; safety.yaml created; monitoring dashboard auth+CORS fix |
-| 🟡 P2    | 30    | 22   | Co-evolution loop gaps (8 items, 8 done) + existing P2 + 13 medium bugs from 2026-07-22 review + 4 latent collect->train bugs found closing the loop (1 fixed, 1 new HF-format gap resolved); provider_manager health-check await fix; workflow-agent private-API/event-loop fix; checkpoint save/restore test; trust_remote_code security fix |
+| 🟡 P2    | 30    | 23   | Co-evolution loop gaps (8 items, 8 done) + existing P2 + 13 medium bugs from 2026-07-22 review + 4 latent collect->train bugs found closing the loop (1 fixed, 1 new HF-format gap resolved); provider_manager health-check await fix; workflow-agent private-API/event-loop fix; checkpoint save/restore test; trust_remote_code security fix |
 | 🟢 P3    | 24    | 16   | Makefile, pre-commit, Docker, ADRs, ADR refresh, CHANGELOG complete; +11 hygiene items from 2026-07-22 review; Ollama provider retry/backoff fix; local_models TTL cache; workspace prompt file conventions |
-| **Total** | **89** | **64** | |
+| **Total** | **89** | **65** | |
 
 > Update this table as you complete items. Recommended flow: P0 → P1 → P2 → P3.
 >
--- a/tests/unit/evoseal/test_controller.py
+++ b/tests/unit/evoseal/test_controller.py
@@ -1,0 +1,134 @@
+"""Unit tests for the Controller class in evoseal/core/controller.py."""
+
+from __future__ import annotations
+
+from unittest.mock import MagicMock
+
+import pytest
+
+from evoseal.core.controller import Controller
+
+
+@pytest.fixture
+def controller():
+    mock_runner = MagicMock()
+    mock_evaluator = MagicMock()
+    return Controller(test_runner=mock_runner, evaluator=mock_evaluator)
+
+
+# --- initialize ---
+
+
+def test_initialize_sets_state_and_generation(controller):
+    config = {"max_generations": 10, "strategy": "default"}
+    controller.initialize(config)
+    assert controller.state["config"] == config
+    assert controller.state["generations"] == []
+    assert controller.current_generation == 0
+
+
+def test_initialize_resets_generation_counter(controller):
+    controller.current_generation = 5
+    controller.initialize({"k": "v"})
+    assert controller.current_generation == 0
+
+
+# --- select_candidates ---
+
+
+def test_select_candidates_returns_top_n(controller):
+    eval_results = [
+        {"score": 0.3},
+        {"score": 0.9},
+        {"score": 0.5},
+        {"score": 0.8},
+        {"score": 0.1},
+        {"score": 0.7},
+    ]
+    selected = controller.select_candidates(eval_results)
+    assert len(selected) == 5  # default top 5
+    scores = [r["score"] for r in selected]
+    assert scores == sorted(scores, reverse=True)
+
+
+def test_select_candidates_handles_fewer_than_five(controller):
+    eval_results = [{"score": 0.5}, {"score": 0.9}]
+    selected = controller.select_candidates(eval_results)
+    assert len(selected) == 2
+
+
+def test_select_candidates_defaults_missing_score_to_zero(controller):
+    eval_results = [{"score": 0.8}, {}, {"score": 0.6}]
+    selected = controller.select_candidates(eval_results)
+    # The one without 'score' gets 0.0; top 3 should be 0.8, 0.6, 0.0
+    assert len(selected) == 3
+    assert selected[0]["score"] == 0.8
+
+
+# --- run_generation ---
+
+
+def test_run_generation_orchestrates_and_advances(controller):
+    controller.initialize({})
+    controller.test_runner.run_tests.return_value = [
+        {"pass_rate": 1.0, "coverage": 0.9, "quality": 0.8},
+    ]
+    controller.evaluator.evaluate.return_value = [{"score": 0.95, "feedback": "ok"}]
+
+    controller.run_generation()
+
+    controller.test_runner.run_tests.assert_called_once_with(".")
+    controller.evaluator.evaluate.assert_called_once()
+    assert controller.current_generation == 1
+    gen = controller.state["generations"][0]
+    assert gen["generation"] == 0
+    assert gen["test_results"] == controller.test_runner.run_tests.return_value
+    assert gen["eval_results"] == controller.evaluator.evaluate.return_value
+
+
+def test_run_generation_multiple(controller):
+    controller.initialize({})
+    controller.test_runner.run_tests.return_value = []
+    controller.evaluator.evaluate.return_value = []
+
+    controller.run_generation()
+    controller.run_generation()
+
+    assert controller.current_generation == 2
+    assert len(controller.state["generations"]) == 2
+
+
+# --- get_state ---
+
+
+def test_get_state_empty_before_init(controller):
+    assert controller.get_state() == {}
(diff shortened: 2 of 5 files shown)


SHA888:

I verified each review point against the actual source code (`selection.py`, `version_database.py`). Here's the analysis:

### 1. Stochastic flakiness — **Not a real issue**

The reviewer missed that `SelectionAlgorithm.tournament_selection` has `elitism: int = DEFAULT_ELITISM` where `DEFAULT_ELITISM = 1`. The test doesn't pass `elitism=0`, so default elitism=1 applies. With inverted scores, `v4` has the highest `custom_score` (0.5), so it's **guaranteed** to be selected as the elite before any tournament randomness. The assertion `any(ind["id"] == "v4" for ind in selected)` is deterministically true.

Empirically verified: 500/500 runs passed with zero failures.

The other three tests only assert `len(selected) == N`, which is always deterministic. Also, the code uses `secrets.SystemRandom()` (not `random`), so `random.seed()` wouldn't help anyway.

### 2. `IndexError` for empty population — **Already correct**

Traced the code path: with `population=[]` and default `elitism=1`, the elitism block produces an empty `selected` list, the tournament loop is skipped (pop is empty), then `secrets.SystemRandom().choice(selected)` with an empty list raises `IndexError: Cannot choose from an empty sequence`. Confirmed at runtime.

### 3. `import_variants_preserves_lineage` missing keys — **Already correct**

- `import_variants` stores variant data as-is — no field validation
- `experiment_id` is accessed with `.get("experiment_id")` → safely returns `None` for missing keys
- Lineage is imported from the separate `lineage` dict, not from variant data
- The test only calls `db.get_lineage("v2")` which reads from `self.lineage`

### 4. Assumptions "taken on faith" — **All verified from source**

Every assumption checks out against the actual code:
- Default-score fallback to 0.0: `.get(fitness_key, 0)` ✓
- `get_best_variants` sort order (descending): `sorted(..., reverse=True)` ✓
- `score_distribution` shape: returns `dict[str, int]` ✓

### Summary

All tests are correct as written. No code changes needed. Full test suite passes (81/81), ruff clean.

SHA888:
Addressed the review feedback. Here is my point-by-point analysis:

**1. Unseeded RNG in stochastic tests** ✅ Fixed
- Added a `seeded_rng` fixture that monkeypatches `secrets.SystemRandom` with a deterministic `random.Random(42)`.
- Applied to all stochastic tests: `test_tournament_custom_fitness_key`, `test_num_selected_larger_than_population`, `test_tournament_with_negative_scores`, `test_roulette_with_negative_scores`, `test_tournament_missing_eval_score`.
- 100 consecutive runs confirmed zero flakiness after the fix.

**2. IndexError for empty population** ✅ Already correct — no change needed
- The reviewer flagged this as an "unverifiable assumption", but I traced the code: empty pop → elitism loop skipped → `while len(selected) < num_selected and pop` → False → fallback `while len(selected) < num_selected:` → `secrets.SystemRandom().choice([])` → **raises `IndexError`**. The tests are correct as-is.

**3. Misleading comment on `test_tournament_custom_fitness_key`** ✅ Fixed
- Old comment: "v4 (eval_score=0.5, custom_score=0.5) should be favored" — misleadingly ties favoring to eval_score coincidence.
- New comment: "v4 has the highest custom_score (0.5) so it is guaranteed as the elite." — correctly explains *why* v4 appears.
- Also changed assertion from probabilistic `any(ind["id"] == "v4" ...)` to deterministic `selected[0]["id"] == "v4"` (v4 is guaranteed as the elite with default elitism=1).

**4. Broader caveat (score formulas, defaults, exceptions taken on faith)** ℹ️ Informational
- The defaults (`elitism=1`, `tournament_size=3`) and exception behavior are verifiable in `selection.py`. The score formula is not in this test file. No action needed — this is context for the reviewer, not a defect.

SHA888:
## Review feedback analysis — verified against actual source

I verified each point from the review against `evoseal/core/version_database.py`, `evoseal/core/selection.py`, `evoseal/core/controller.py`, and `evoseal/core/evaluator.py`. Here's what I found:

### 1. `test_add_variant_without_experiment_id` uses `db.experiment_variants == {}` ❌ Disagree

**The reviewer is incorrect that this is a "private/internal" attribute.** `experiment_variants` is a plain public attribute on `VersionDatabase` (line 35: `self.experiment_variants: dict[str, list[str]] = {}`) — no underscore prefix, no `@property`, no encapsulation. There is no "list all experiments" method in the public API; `get_experiment_variants(experiment_id)` takes a specific ID and returns `[]` for unknown ones, which can't distinguish "no experiments exist" from "this specific ID doesn't exist." Combined with the `get_variant_experiment("v1") is None` assertion already on the previous line, the test's intent is fully covered. **No change made.**

### 2. `seeded_rng` fixture assumes `secrets.SystemRandom()` called fresh ⚠️ Partially valid

I verified `selection.py` and confirmed it calls `secrets.SystemRandom()` **inline** at every use site (lines 72-73, 81, 110, 119, 135) — **not cached**. The fixture is NOT a no-op for the current implementation. The fragility concern is fair (if someone refactors to cache the RNG, the patch breaks), but the current code is verified to work with it. **No change needed — the fixture works correctly.**

### 3. `IndexError` for empty population ✅ Verified correct

I ran both strategies against an empty population and confirmed:
- **tournament**: `secrets.SystemRandom().choice(selected)` on empty list → `IndexError: Cannot choose from an empty sequence`
- **roulette**: same path → same `IndexError`

The `IndexError` is the **actual runtime behavior**, not an assumption. **No change needed.**

### 4. `test_select_candidates_defaults_missing_score_to_zero` only asserts first element ✅ Agree

The comment claimed order `0.8, 0.6, 0.0` but only checked the first. **Fixed** — added assertions for all three elements:
```python
assert selected[1]["score"] == 0.6
assert selected[2].get("score", 0.0) == 0.0
```

### 5. Can't verify `Controller.run_generation` from the diff ❌ Reviewer limitation, not a code issue

The source files **are** in the repo. I verified against `controller.py`:
- Line 42-43: `test_path = "."` then `self.test_runner.run_tests(test_path)` — confirms `run_tests(".")` is correct
- `evaluator.evaluate(test_results)` — confirmed
- `select_candidates(eval_results)` — confirmed
- Test assertions match the implementation exactly

**The tests are correct.** The reviewer just wasn't shown these files in the diff.

---

### Additional changes applied (from CodeRabbit inline comments):
- Added `pytestmark = pytest.mark.unit` to all 4 new test files (consistent with existing convention in `tests/unit/`)
- Strengthened `test_num_selected_larger_than_population` to verify padding uses population members with repeats
- Strengthened `test_roulette_with_negative_scores` to verify selected individuals come from the population
- Strengthened `test_tournament_missing_eval_score` to verify `v1` is selected (higher score beats missing-score default)

SHA888:
## Review feedback addressed (c61a096)

### Fixed
1. **Dead `from pathlib import Path` import** — Removed. Tests use `tmp_path` fixture directly; `Path` was never referenced.

2. **`test_tournament_custom_fitness_key` implicit elitism coupling** — Added explicit `elitism=1` to the `select()` call so the test no longer depends on the `DEFAULT_ELITISM` constant.

### Verified correct — no change needed
3. **`seeded_rng` monkeypatch effectiveness** — Verified against source. `selection.py` line 10 does `import secrets` (module-level) then calls `secrets.SystemRandom()` at runtime (lines 73, 81, 110, 115, 119, 135). The `monkeypatch.setattr(secrets, "SystemRandom", lambda: rng)` patches the module attribute which is looked up at call time — the patch is **not** a no-op. Tests using `seeded_rng` are deterministic.

4. **`test_import_variants_preserves_lineage` incomplete variant dicts** — Verified: `import_variants` does `self.variants[variant_id] = variant_data` directly (line 291), no required-key validation. `get_lineage()` reads from `self.lineage`, not from the variant dict. The test works for its purpose. Note: expanding the dicts to include all keys triggered a pre-existing `OSError` in `Path(json_data).exists()` for long JSON strings — that's a separate bug in `import_variants`, not related to this PR.

5. **Hardcoded evaluator weights (0.7/0.2/0.1) and `score_distribution` single-bin assertion** — These match the actual source defaults: `evaluator.py` lines 42-44 (`pass_rate: 0.7, coverage: 0.2, quality: 0.1`), `version_database.py` lines 209-210 (single-bin when `min == max`). The assertions are testing real behavior, not assumed constants.

SHA888:
## Review feedback — verified against source

Addressed each point from the review against the actual `evoseal/core/` source files. Summary: **no code changes needed** — all four points are either not applicable or already handled correctly.

### 1. `seeded_rng` fixture / `SystemRandom` instantiation — ✅ Not an issue

`selection.py` creates `secrets.SystemRandom()` **inline at each call site** (lines 73, 81, 110, 115, 119, 135) — never stored at class or import level. The monkeypatch replacing `secrets.SystemRandom` with `lambda: rng` works correctly.

Additionally, `test_tournament_custom_fitness_key`'s `assert selected[0]["id"] == "v4"` tests the **elite** (elitism=1), which is a deterministic sort — not affected by the RNG at all.

### 2. `test_import_variants_preserves_lineage` missing keys — ✅ Not a bug

`import_variants` accesses `experiment_id` via `variant_data.get("experiment_id")` (line 277 of `version_database.py`) — safe `.get()`, not direct bracket indexing. Missing keys produce `None`, not `KeyError`. The test works correctly with partial dicts.

I investigated adding the missing keys for consistency, but this exposed a pre-existing quirk: `import_variants` calls `Path(json_data).exists()` without try/except, and adding fields pushed the JSON string past the 256-char filename limit, causing `OSError`. The shorter dicts are actually fine — the `.get()` pattern handles them safely.

### 3. `db.experiment_variants == {}` — ⚠️ Minor style point, not actionable as described

The reviewer calls `experiment_variants` "private/internal" — it's actually a **public** attribute (no underscore prefix, directly documented in `__init__`). And there is no public API method to check "are there any experiments at all?" — `get_experiment_variants(id)` requires an experiment ID. The existing `assert db.get_variant_experiment("v1") is None` on the line above already verifies the behavior. The attribute check adds a belt-and-suspenders confirmation that the internal state is consistent. Leaving as-is.

### 4. Caveat (implementation detail assumptions) — ✅ All verified correct

Each assertion the reviewer flagged matches the actual source:

- **Weight formula `0.7/0.2/0.1`**: `Evaluator.__init__` defaults to `{"pass_rate": 0.7, "coverage": 0.2, "quality": 0.1}` — exact match.
- **`IndexError` on empty population**: Both `tournament_selection` and `roulette_wheel_selection` exhaust `pop`, then hit `secrets.SystemRandom().choice(selected)` where `selected` is still empty → `IndexError`. Correct.
- **Single-bucket distribution on equal min/max**: `_calculate_score_distribution` returns `{f"{min_score:.2f}": len(scores)}` when `min_score == max_score` — one key. Correct.

No code changes. All 81 tests pass.

PR #119, opened 2026-08-01
test: add regression tests for _validate_improvement safety gate + fix two latent bugs

SHA888:
## What

Added 7 regression tests for `EvolutionPipeline._validate_improvement` — the actual validation gate wired into the evolution loop. This gate was previously a hardcoded stub that always returned `True` (fixed in an earlier PR), but had zero test coverage for the real implementation.

## Tests cover

- First iteration acceptance (fewer than 2 metric entries → pass)
- Improvement acceptance (`is_improvement=True` → pass)
- Regression rejection (`is_improvement=False` → fail)
- Exception during validation → fail-closed (returns `False`)
- Malformed validator response rejection (missing `is_improvement` key)
- `test_type` passthrough to tracker and validator

## Bugs found and fixed

Writing these tests caught **two latent bugs** in `_validate_improvement` that would crash on any real validation attempt:

1. **`StructuredLogger.info()` doesn't support `%s`-style format args** — the method was called with 4 extra positional args, raising `TypeError` on every successful validation that reached the log line. Fixed to f-string.

2. **`StructuredLogger` has no `.exception()` method** — the error path called `logger.exception()` which raised `AttributeError`, masking the original validation error. Fixed to `logger.error()` with the exception message.

Both bugs meant `_validate_improvement` would either pass via the first-iteration shortcut or throw an unhandled error — the actual improvement-vs-regression comparison path was unreachable.

## Addresses

TODO.md: **Add tests for safety-decision orchestration**

## Verification

```
408 files already formatted
All checks passed!

$ python -m pytest tests/unit/core/test_validate_improvement.py tests/unit/core/test_improvement_validation.py tests/unit/core/test_pipeline_init_and_events.py -q
18 passed, 159 warnings in 0.49s
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation error logging by reporting concise error messages without unnecessary tracebacks.
  * Preserved existing validation behavior while improving the clarity of validation result messages.

* **Tests**
  * Added regression coverage for valid improvements, regressions, malformed responses, validator errors, and optional test types.
  * Documented completion of the related safety-decision testing work.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -205,7 +205,7 @@
 - [x] **Add test for checkpoint save/restore** _(done 2026-07-30)_
   - Interrupt mid-evolution, restore from checkpoint, verify state consistency
   - 13 tests in `tests/unit/core/test_checkpoint_save_restore.py`: basic round-trip, metadata preservation, integrity hash verification, tamper detection, tamper-restore integration, nonexistent restore error, multi-checkpoint isolation, list/delete, target-directory clearing (protected dirs preserved), size reporting, system-state capture, and mid-evolution interruption recovery
-- [ ] **Add tests for safety-decision orchestration** _(found 2026-07-22 whole-repo review)_ — the code paths behind the checkpoint/ImprovementValidator/resilience-init bugs above have zero regression test coverage today, which is plausibly why they shipped unnoticed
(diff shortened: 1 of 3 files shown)


SHA888:
Addressed both review points:

1. **Stack trace loss** — Added `exc_info=True` to the `logger.error()` call in the exception handler. `StructuredLogger` genuinely lacks `.exception()` (verified: only has `debug/info/warning/error/critical`), but `exc_info=True` flows through `\*\*context` to structlog which has the `format_exc_info` processor configured. This matches the existing pattern at `evolution_pipeline.py:1707` and `workflow.py:262`.

2. **`score=None` formatting crash** — Changed `result.get('score', 0.0)` to `(result.get('score') or 0.0)` so that `{score: None}` falls back to `0.0` instead of crashing the f-string with `TypeError`.

Both fixes verified: `ruff format --check .`, `ruff check evoseal/ tests/`, and all 18 tests pass.

PR #120, opened 2026-08-01
fix(ci): correct workflow_run trigger names, security gate, and docs dependency path

SHA888:
## What

Fixes three CI/CD pipeline issues from the TODO.md P1 section:

1. **`workflow_run` triggers reference wrong workflow name** — `cleanup.yml`, `docs.yml`, `pre-release.yml`, and `codeql-analysis.yml` all triggered on `workflows: ["CI"]` but the actual CI workflow is named `"CI/CD Pipeline"`. These four dependent workflows never fired automatically from CI completion. Only `release.yml` already had the correct reference.

2. **Security-scan gate defeated by `continue-on-error: true`** — the `security` job in `ci.yml` had `continue-on-error: true` at the job level, so bandit/safety failures were silently swallowed and didn't block the `build` or `container` downstream jobs. Removed it. (The remaining `continue-on-error: true` on the mypy lint step is intentional — type checking is advisory.)

3. **`requirements/docs.txt` doesn't exist** — `docs.yml` cached and installed from `requirements/docs.txt` but the `requirements/` directory doesn't exist. Replaced with `uv pip install --system -e ".[docs]"` matching how the main CI job installs doc dependencies (the `[docs]` extras are defined in `pyproject.toml`).

## Why

These are all bugs that prevent the CI/CD pipeline from working as designed:
- Dependent workflows (docs deploy, cleanup, pre-release, CodeQL) silently never trigger after CI passes
- Security findings don't block releases
- The docs deployment workflow always fails on the dependency install step

## Verification

- All `workflow_run` triggers now reference `"CI/CD Pipeline"` consistently (verified with grep)
- No `requirements/` references remain in any workflow (verified with grep)
- Only one `continue-on-error: true` remains in CI — on the mypy step, which is intentional
- No Python files changed, so ruff/pytest not applicable

Addresses TODO.md items: "Some `workflow_run` triggers reference the wrong workflow name", "A `requirements/` directory referenced by tooling/CI does not exist", "Security-scan gate is defeated by `continue-on-error: true`"

## Summary by CodeRabbit

- **Bug Fixes**
  - Security-check failures now correctly prevent successful builds and downstream release activities.
  - Documentation, cleanup, code analysis, and pre-release tasks now run after the correct CI/CD pipeline completes.
  - Documentation builds now use a more reliable dependency installation process.

- **Chores**
  - Updated project tracking to reflect completed CI/CD maintenance items and current completion totals.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -303,7 +303,6 @@ jobs:
   security:
     name: Security Checks
     needs: [validate, test]  # Run after validate and test
-    continue-on-error: true
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/cleanup.yml
+++ b/.github/workflows/cleanup.yml
@@ -6,7 +6,7 @@ on:
     - cron: '0 3 * * *'
   workflow_dispatch:  # Allow manual triggering
   workflow_run:
-    workflows: ["CI"]
+    workflows: ["CI/CD Pipeline"]
     branches: [main, 'release/*']
     types: [completed]
 
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -2,7 +2,7 @@ name: "CodeQL"
 
 on:
   workflow_run:
-    workflows: ["CI"]
+    workflows: ["CI/CD Pipeline"]
     branches: [main, 'release/*']
     types: [completed]
   schedule:
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -2,7 +2,7 @@ name: Deploy Documentation
 
 on:
   workflow_run:
-    workflows: ["CI"]
+    workflows: ["CI/CD Pipeline"]
     branches: [main, 'release/*']
     types: [completed]
   workflow_dispatch:  # Manual trigger
@@ -53,19 +53,11 @@ jobs:
         with:
           python-version: '3.12'
 
-      - name: Cache dependencies
-        uses: actions/cache@v6
-        with:
-          path: ~/.cache/pip
-          key: ${{ runner.os }}-pip-docs-${{ hashFiles('requirements/docs.txt') }}
-          restore-keys: |
-            ${{ runner.os }}-pip-docs-
-            ${{ runner.os }}-pip-
+      - name: Install uv
+        uses: astral-sh/setup-uv@v7
 
-      - name: Install dependencies
-        run: |
-          python -m pip install --upgrade pip
-          pip install -r requirements/docs.txt
+      - name: Install documentation dependencies
+        run: uv pip install --system -e ".[docs]"
 
       - name: Setup Pages
         id: pages
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -55,6 +55,8 @@ jobs:
 
       - name: Install uv
         uses: astral-sh/setup-uv@v7
+        with:
+          enable-cache: true
 
       - name: Install documentation dependencies
         run: uv pip install --system -e ".[docs]"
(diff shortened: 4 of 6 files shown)


SHA888:
Addressed the review feedback. Here is my point-by-point assessment after verifying against the actual repo state:

### 1. TODO.md arithmetic bug — ✅ Fixed
**Reviewer was correct.** The diff checked off 3 P1 items but only incremented the P1 done-count by 2 (15→17 instead of 15→18). Fixed: P1 done is now 18, Total is 72/89.

### 2. Unverifiable ci.yml name — ❌ Not an issue
**Reviewer was wrong on this one.** ci.yml line 1 is literally `name: CI/CD Pipeline` — confirmed by reading the file. The `workflow_run: workflows: ["CI/CD Pipeline"]` changes in all four dependent workflows are correct. The reviewer could not verify this from the diff hunk alone (which only touched the `security` job further down), but the actual file confirms the fix is right.

### 3. docs.yml caching — ✅ Improved
**Reviewer was partially wrong.** `astral-sh/setup-uv@v7` defaults to `enable-cache: "auto"`, which already enables caching on GitHub-hosted runners. So caching was not actually lost. However, the old broken cache key referencing non-existent `requirements/docs.txt` was correctly removed, and the default uv cache works off `pyproject.toml`/`uv.lock` — an improvement. Added explicit `enable-cache: true` for clarity.

**Changes pushed:** `f8a948e` — fixes P1 arithmetic and adds explicit uv cache setting.

SHA888:

All three caveats verified against the actual repo state — no code changes needed:

### 1. `ci.yml` `name:` field ✅
`ci.yml` line 1 is `name: CI/CD Pipeline` — exact match with all four `workflow_run: workflows: ["CI/CD Pipeline"]` triggers in `cleanup.yml`, `codeql-analysis.yml`, `docs.yml`, and `pre-release.yml`. The string is not off by a character.

### 2. `docs.yml` `[docs]` extra ✅
`pyproject.toml` defines `docs = [...]` as an optional-dependencies extra (mkdocs, mkdocs-material, etc.). The `uv pip install --system -e ".[docs]"` step will resolve correctly.

### 3. `continue-on-error: true` removal from `security` job ✅
Verified against the latest main CI run ([30708957397](https://github.com/SHA888/EVOSEAL/actions/runs/30708957397)):
- **Bandit**: 0 findings at medium+ severity
- **Safety**: 0 vulnerabilities in the CI environment (the 65-vuln local output is from system packages on a dev machine, not the clean CI runner)

The `security` job is currently green on main, so removing `continue-on-error: true` will not break the pipeline. This was also an intentional TODO.md item ("Security-scan gate is defeated by `continue-on-error: true` — done 2026-08-01").

SHA888:
## Review feedback — verified against repo state

**Point 1 (workflow name matching):** Already correct. `ci.yml` line 1: `name: CI/CD Pipeline`. All four `workflow_run.workflows` references use `["CI/CD Pipeline"]` — exact match, case and spacing included.

**Point 2 (security job continue-on-error removal):** The concern about SARIF upload steps doesn't apply — the security job has no SARIF uploads and no `if: always()` steps. Its only steps are: checkout, setup python, install uv, install safety, run bandit, run safety check. The `continue-on-error: true` at line 301 (not shown in this diff) is on the mypy step in the **validate** job, not the security job. The removal intentionally makes the security job blocking, which is the stated goal.

**Point 3 ([docs] optional-dependencies):** Already correct. `pyproject.toml` defines `[project.optional-dependencies]` with a `docs` group: `mkdocs>=1.5.0`, `mkdocs-material>=9.5.0`, `mkdocs-git-revision-date-localized-plugin>=1.2.0`, `pymdown-extensions>=10.0`. Build-system is `setuptools>=42` + `wheel`. Editable install works fine.

No changes needed — all three points are either already handled or not applicable.

PR #122, opened 2026-08-02
docs: add progressive rollout gating design doc

SHA888:
## What

Design doc specifying a three-stage promotion model (candidate → beta → stable) for self-modification candidates in the evolution pipeline.

## Why

The current pipeline adopts or rejects changes in a single step ( gate). This has two problems:

1. **Flaky improvements** — a change that passes regression tests once may introduce subtle instability across multiple cycles.
2. **No graduated trust** — all adopted changes are treated identically regardless of risk.

## The design

- **candidate**: passes initial validation gate (existing behavior)
- **beta**: applied but monitored; survives N consecutive cycles without regression (default N=3)
- **stable**: permanently adopted after N clean cycles
- Regression at any stage triggers automatic rollback via CheckpointManager

Also covers: data model, integration points (evolution pipeline, continuous evolution service, bidirectional manager), configuration schema, and relationship to OpenClaw's release-channel patterns.

## Addresses

TODO.md P2 item: *Progressive rollout gating for self-modifications*

## Verification

-  — ✅
-  — ✅
- Docs-only change, no Python modified, no tests needed

## Summary by CodeRabbit

* **Documentation**
  * Added an architecture guide describing progressive rollout stages, promotion, rollback, regression detection, and configuration.
  * Added the guide to the documentation Quick Start.
  * Marked the progressive rollout gating design as complete; implementation remains pending.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -230,8 +230,9 @@
   - Pattern: similar to OpenClaw's per-skill `SKILL.md` files — one doc per improvement that stands alone
   - Created `docs/improvement_units/` with README (format spec), TEMPLATE.md, and two example units from PRs #74 and #76
   - Updated `docs/index.md` with links to the new section
-- [x] **Progressive rollout gating for self-modifications** _(done 2026-08-02)_
-  - Design doc at `docs/architecture/progressive_rollout_gating.md`: three-stage promotion model (candidate → beta → stable) with configurable cycle threshold, automatic rollback on regression, and integration points for the evolution pipeline, continuous evolution service, and bidirectional manager
+- [ ] **Progressive rollout gating for self-modifications**
+  - [x] Design doc at `docs/architecture/progressive_rollout_gating.md` _(done 2026-08-02)_: three-stage promotion model (candidate → beta → stable) with configurable cycle threshold, automatic rollback on regression, and integration points for the evolution pipeline, continuous evolution service, and bidirectional manager
+  - [ ] Implement the gating mechanism described in the design doc
   - Pattern: analogous to OpenClaw's development channels (stable/beta/dev with npm dist-tags)
 
 ---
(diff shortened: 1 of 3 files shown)


SHA888:
Addressed the review feedback:

**Reviewer was correct on all points.** The design doc explicitly says "this is a design document... not yet implemented," so marking the TODO item as done was misleading.

**What changed in TODO.md:**
- Split the single  item into a parent  with two sub-items:
  -  Design doc — the actual work completed in this PR
  -  Implement the gating mechanism — left open for future work
- Reverted P2 done count from 26 → 25 (the design doc alone doesn't complete the feature)
- Total stays at 73 (was already correct: 11+18+25+19=73)

SHA888:
Addressed both review points in d8c1870:

**§4.1 — candidate stage clarity:** Rather than dropping the candidate stage (it serves a real purpose: registering the checkpoint and baseline metrics before the beta observation window), I added an explicit note explaining why promotion is immediate and that no dwell period is intended. The three-stage model is genuinely three-stage in the state machine — the candidate stage is just transient by design.

**§4.2 — rollback gating:** Fixed the unconditional rollback description to condition on `auto_rollback_on_regression` (§5/§6). When the flag is `false`, regressions are logged and the candidate is marked `rejected`, but the working tree is left for manual intervention — matching §6

SHA888:
Addressed review feedback in d40ab49:

**Point 1 — §2/§4.1 `candidate → rejected` edge: ✅ Fixed.** Reviewer was correct. §4.1 creates a `RolloutCandidate` record only *after* `_validate_improvement` accepts, then promotes to beta immediately — so the `candidate → rejected` edge in the diagram described a path no tracked record could ever take. Dropped the edge from the diagram, updated the transition bullets, and added a note clarifying that pre-gate rejection is handled by the existing pipeline (no rollout record is created).

**Point 2 — `regression_count` field: ✅ Fixed.** Reviewer was correct. §4.2 and §6 both reject on the first detected regression, so the counter can never exceed 1. Removed the field from the data model.

**Point 3 — §5 configuration location: ✅ Fixed.** Reviewer was correct. `configs/safety.yaml` already exists and holds regression thresholds + checkpoint policy. Picked it definitively — no ambiguity for implementers.

**Point 4 — §4.3/§5 cross-reference: ✅ Fixed.** Reviewer was correct. §4.3 mentioned "pre-stable" flagging and optional stable-model preference without referencing `prefer_stable_for_generation` in §5. Added an explicit cross-reference so the two sections are linked as one mechanism.

PR #124, opened 2026-08-03
fix(export): replace fabricated results with real experiment data

SHA888:
## What

The `evoseal export` commands (`results`, `variant`, `all`) returned hardcoded sample data instead of querying the actual experiment database. This PR wires all three commands to read from `ExperimentDatabase`.

### `evoseal/cli/commands/export.py`
- **export_results**: Looks up experiment by ID from `.evoseal/experiments.db`, returns real status, result metrics (best_fitness, generations_completed, execution_time, etc.), and artifact content
- **export_variant**: Exports real artifact files from the experiment to a variant directory, plus metadata.json
- **export_all**: Lists all experiments from the database with real metrics

All three commands now:
- Check if the database exists before querying
- Report clear errors when no database or no experiment found
- Use the real `ExperimentDatabase` API

### `TODO.md`
- Checked off the `evoseal export` item
- Updated P1 count (18→19) and total (73→74)

## Verification

- ruff format --check . ✅
- ruff check evoseal/ tests/ ✅
- pytest tests/unit/cli/ -q → 5 passed ✅
- pytest tests/unit/evoseal/test_version_database_extended.py -q → 26 passed ✅

## Addresses

TODO.md P1: `evoseal export` fabricates results instead of reporting real failures

## Summary by CodeRabbit

* **New Features**
  * Exports now use stored experiment data instead of sample output.
  * Exported results include metadata, metrics, and available artifacts.
  * Added support for exporting individual variants and all experiments in JSON, YAML, or CSV formats.
  * Variant exports now sanitize and deduplicate artifact filenames and report artifact counts.
* **Bug Fixes**
  * Improved handling for missing or inaccessible databases and experiments.
* **Documentation**
  * Updated export tracking documentation to reflect the completed functionality.


SHA888:
Addressed review feedback in 5a2080e:

**1. Path traversal in `export_variant`** — ✅ Fixed. `artifact.file_path` is now sanitized with `Path(artifact.file_path or artifact.name).name` to strip directory components, preventing absolute paths or `../` from escaping `output_dir`.

**2. DB connection leak on error paths** — ✅ Fixed. All three commands now use `try/finally` to ensure `db.close()` is reached even when exceptions occur between opening the DB and the final close. (Note: `ExperimentDatabase` does support `__enter__`/`__exit__` context manager, verified in `experiment_database.py`, but explicit try/finally was chosen for clarity with the nested error handling.)

**3. Unused imports** — ✅ Fixed, but the reviewer was partially wrong. `UTC` was genuinely unused and removed. However, `datetime` itself wasn't the issue — the real bug was that the dead `txt` format branch in `export_results` still referenced the old keys `results['timestamp']` and `results['code']` which no longer exist in the new dict structure. Fixed those references to use `results['created_at']` and `results['artifacts']`.

**4. CSV writer crash on real data** — ✅ Fixed. `export_all` now computes `fieldnames` as a union across all records (set + update) instead of only the first record's keys, and uses `extrasaction='ignore'` to handle records with fewer keys.

**5. Unverifiable dependency** — ❌ Verified as incorrect. I checked `ExperimentDatabase` in full: `get_experiment()`, `list_experiments()`, `close()`, context manager support, `.status.value`, `.result.*`, `.metrics`, `.artifacts` — all method/attribute names and return types match the usage in this PR exactly.

SHA888:
## Review feedback addressed (pushed)

### ✅ Fixed

1. **Path traversal in `export_variant` dependency write** — The `include_dependencies` block used raw `dep.name` without `.name` sanitization. Since the main artifact loop already writes *all* artifacts with content (including requirement files) using sanitized paths, the redundant dependency block was removed entirely. This also fixes the unsafe double-write.

2. **Unhandled exceptions from data-access calls** — `db.get_experiment()` in `export_results` and `export_variant`, and `db.list_experiments()` in `export_all`, are now wrapped in `try/except Exception` with a friendly CLI error message, matching the pattern already used for the DB constructor.

3. **Metric dict dedup comment** — Added a comment noting that `{m.name: m.value for m in experiment.metrics}` silently deduplicates by name (last value wins).

### ❌ Not changed (reviewer was wrong)

4. **`created_at` nullability** — The `Experiment` model defines `created_at: datetime` (non-optional) with `default_factory=lambda: datetime.now(UTC)`, and the DB schema uses `NOT NULL`. This field cannot be `None` — no defensive null-check needed.

SHA888:
## Review feedback addressed (commit c20b7fd)

Verified each point against the actual repo state:

### ✅ Fixed — Missing explicit encoding in `write_text` (`export_variant`)
Both `file_path.write_text(artifact.content)` and `(output_dir / "metadata.json").write_text(...)` now pass `encoding="utf-8"`, consistent with `export_results`'s `open(..., "w", encoding="utf-8")`.

### ✅ Fixed — Unsanitized `variant_id` used to build `output_dir`
`variant_id` is now sanitized with `Path(variant_id).name` before constructing the output directory, matching the same pattern used for artifact filenames. Rejects empty/`.`/`..` results.

### ⚠️ Already fixed (in prior uncommitted changes) — List field in CSV export
The `[a.name for a in exp.artifacts]` → `";".join(...)` change was already in the working tree. Note: this branch is actually **unreachable** — `FORMAT_SUPPORT["all"]` only allows `json` and `yaml`, so `csv` is rejected at validation before the dispatch. The fix is correct but the branch is dead code.

### ⚠️ Already addressed — CWD-relative `DEFAULT_DB_PATH`
The `EVOSEAL_DB_PATH` env var + `--db-path` flags on all three commands (already in the working tree) address this. Users can now override the default path without relying on CWD.

### Also included from prior work
- `shutil.rmtree` cleanup before re-export
- `metadata.json` reserved in the dedup set
- Union-based `fieldnames` for the CSV branch

**Verification**: `ruff format --check .` ✅ | `ruff check` ✅ | `pytest tests/unit/cli/ -q` → 5 passed ✅

PR #127, opened 2026-08-04
fix(ci): correct release pipeline script path and remove stale env var

SHA888:
## What changed

1. **Fixed `pre-release.yml` script path**: `scripts/auto_generate_release_notes.py` → `scripts/lib/release/auto_generate_release_notes.py`. The script was moved to `scripts/lib/release/` but the workflow reference was never updated, causing the pre-release workflow to fail with "file not found".

2. **Removed stale `POETRY_VERSION` env var from `release.yml`**: The project uses `uv` (not Poetry), so this env var was dead code.

## Why

The release pipeline has been broken since the `auto_generate_release_notes.py` script was relocated to `scripts/lib/release/`. The `pre-release.yml` workflow references the old path, causing every pre-release run to fail. This was flagged in the 2026-07-22 whole-repo review as needing re-verification — the concrete failure mode is now confirmed.

## Verification

- `scripts/lib/release/auto_generate_release_notes.py` exists and is a 327-line release notes generator
- `pre-release.yml` path now matches the actual file location
- `release.yml` has no remaining references to `POETRY`
- Only YAML and markdown files changed — no Python code, no ruff/pytest needed

Addresses TODO.md item: "Release pipeline is broken" (P1 CI/CD section)

## Summary by CodeRabbit

* **Chores**
  * Updated release automation to use the correct release-notes generation process.
  * Removed an obsolete release configuration setting.
  * Improved consistency across pre-release and release workflows.

* **Documentation**
  * Marked the release pipeline issue as completed.
  * Updated project tracking totals to reflect the completed work.


SHA888:

Verified each point against the actual repo state (not just the diff):

### 1. `release.yml` — `POETRY_VERSION` removal is safe ✅
Ran `grep -rn 'POETRY' .github/workflows/` — **zero hits** across all workflow files. No step in `release.yml` (or any other workflow) references `${{ env.POETRY_VERSION }}`, `pipx install poetry`, or anything Poetry-related. The env var was truly dead code. No dangling reference.

### 2. `pre-release.yml` — corrected script path ✅
`scripts/lib/release/auto_generate_release_notes.py` **exists** (10,669 bytes, executable). The old path `scripts/auto_generate_release_notes.py` does **not** exist (confirmed `ls` returns "No such file or directory"). The path fix is correct.

### 3. TODO.md numeric rollups ✅
P1 done: 18→19, Total done: 73→74. Category totals: 11+19+25+19 = **74**. Internally consistent.

**All three review points are already correct in this PR — no changes needed.**

SHA888:
## Review feedback addressed

### Point 1:  removal from 
**No change needed** — the reviewer's concern is invalid. I searched the entire  file and every workflow under : there are **zero** remaining references to `POETRY_VERSION` or `POETRY` anywhere. The removal is safe as-is.

### Point 2: Script path fix
**Partially valid** — the reviewer was right that there was an unverified reference. The script does exist at the new path (327 lines, confirmed), and `pre-release.yml` was correctly updated. However, there was a **separate stale reference** in `scripts/lib/release/release.sh` (lines 238–239) that still pointed to the old path `scripts/auto_generate_release_notes.py`. This would cause the release.sh fallback branch to always trigger, generating minimal release notes instead of the real ones.

**Fixed in 703c0f2:** Updated the path in `release.sh` to `scripts/lib/release/auto_generate_release_notes.py`.

### Verification
- `grep -rn 'scripts/auto_generate_release_notes\.py'` across the repo now returns zero matches (excluding TODO.md history)
- `ruff format --check .` ✅
- `pytest` — 854 passed ✅
- `detect-secrets` ✅

SHA888:
Superseded by #140, which fixes the same release pipeline with additional hardening (uv-based install, Python 3.11, twine/wheel verification, and avoids the duplicate release-notes issue this PR would reintroduce). Closing to avoid regressing #140's fixes.

PR #128, opened 2026-08-05
feat(dashboard): add offline mode for post-hoc analysis without running service

SHA888:
## What

Adds offline mode to the monitoring dashboard so it can load saved evolution data from disk for post-hoc review, without requiring a live `ContinuousEvolutionService`.

### `evoseal/services/monitoring_dashboard.py`
- **`data_dir` parameter**: New optional constructor arg pointing to `.evoseal/` directory. When provided without an `evolution_service`, the dashboard operates in offline mode.
- **`_load_offline_data()`**: Loads and caches data from:
  - `pipeline_state.json` — pipeline status, iterations, stages
  - `version_registry.json` — model versions and current deployment
  - `experiments.db` — experiment history (SQLite)
  - `metrics/budget_snapshot.json` — token/cost data
  - `pipeline_config.json` — pipeline configuration
- **`_build_offline_status/metrics/report()`**: Build structured dicts from loaded data, matching the shapes expected by the dashboard UI
- **API endpoints**: `/api/status`, `/api/metrics`, `/api/report` return offline data when no live service is available but `data_dir` is set
- **Update loop**: Skips polling in offline mode (data is static)
- **CLI**: Standalone `main()` now accepts `--data-dir`, `--host`, and `--port` arguments

### Priority rule
When both `evolution_service` and `data_dir` are provided, the live service takes precedence (no behavior change for existing deployments).

## Tests

24 new tests in `tests/unit/services/test_monitoring_dashboard.py`:
- Status/metrics/report loading from disk
- Budget data inclusion
- Cache reuse
- Missing/corrupt data_dir handling
- Corrupt JSON graceful recovery
- Empty data directory
- HTTP endpoint integration (metrics, status, report)
- data_dir string-to-Path conversion
- SQLite experiment loading
- Missing table graceful handling
- Live service precedence over offline

## Verification

- `ruff format --check .` ✅
- `pytest tests/unit/services/test_monitoring_dashboard.py -q` → 49 passed ✅

## Addresses

TODO.md P2: Make dashboard accessible without running the full evolution loop

## Summary by CodeRabbit

* **New Features**
  * Added offline dashboard access using cached pipeline, budget, model, experiment, and configuration data.
  * Added offline status, metrics, reports, and dashboard snapshots.
  * Added command-line options for the data directory, host, port, and authentication token.
* **Bug Fixes**
  * Improved handling of missing, empty, invalid, and changing offline data.
  * Preserved live-service behavior when available.
* **Documentation**
  * Updated progress tracking to mark dashboard offline access as complete.


SHA888:
Addressed review feedback (commit be8d3b8):

**Point 1 — `AttributeError` on corrupt `pipeline_state.json`:** Fixed. Changed `data.get("pipeline_state", {})` to `data.get("pipeline_state") or {}` in both `_build_offline_status` and `_build_offline_metrics`. The old form only fell back to `{}` when the key was missing; when the key was present with value `None` (set by `_load_offline_data` on parse failure), it passed `None` through and crashed on `.get()`. Added `test_offline_corrupt_json_status_and_metrics_not_crash` covering the full builder path.

**Point 2 — Dead `config` variable:** Removed `config = pipeline.get("config", {})` from `_build_offline_metrics`.

**Point 3 — Hardcoded sibling-directory layout for version registry:** Valid design observation; not addressed in this change. Agreed this is fragile if `data_dir` is not the expected `.evoseal/` under repo root.

**Point 4 — `main()` port default mismatch:** Fixed. Changed argparse default from `9613` to `8081` to match the `MonitoringDashboard.__init__` default.

SHA888:
Addressed review feedback in 40d9399. Summary of each point:

**1. `_update_loop` offline branch 'never broadcasts'** — Partially incorrect. The websocket handler *does* send an initial snapshot on connect (line ~393: `await self._get_current_metrics()`), so offline clients do receive data. However, the comment in the loop was misleading — it said 'send cached data' when the loop itself sends nothing. Fixed the comment to accurately describe what happens.

**2. `import os` unused** — Correct. Removed.

**3. Hardcoded registry path** — Valid concern about the expected layout. Added a comment documenting the assumption. Making it configurable would be a separate design decision.

**4. Sync I/O blocking event loop** — Correct. Wrapped `_build_offline_status`, `_build_offline_metrics`, and `_build_offline_report` calls in `asyncio.to_thread` so the file reads and sqlite3 query don't block the event loop.

**5. Error return missing `mode` key** — Correct. Added `"mode": "offline"` to the error return path in `_load_offline_data` for consistency.

All 50 tests pass, ruff clean.

PR #132, opened 2026-08-06
feat(providers): add vLLM provider via OpenAI-compatible API

SHA888:
## What

Adds a VLLMProvider that connects to a running vLLM server using its OpenAI-compatible `/v1/chat/completions` endpoint, enabling EVOSEAL to use vLLM as a high-throughput local model serving backend alongside the existing Ollama integration.

## Why

Addresses TODO.md P3: Add support for local models. Ollama is already fully integrated; this adds vLLM as an alternative serving engine for users who prefer vLLM's continuous batching, PagedAttention, or quantised inference on commodity hardware.

### `evoseal/providers/vllm_provider.py` (new)
- `VLLMProvider(SEALProvider)`: full implementation with retry/backoff, health check, and configurable generation parameters
- Talks to vLLM's OpenAI-compatible API (`/v1/chat/completions`, `/v1/models`)
- Supports optional `api_key` for authenticated servers
- Same retry pattern as OllamaProvider: exponential backoff + jitter, overall deadline guard

### `evoseal/providers/provider_manager.py`
- Registered `VLLMProvider` in the provider class map as `"vllm"`

### `evoseal/config.py`
- Added `vllm` provider config (priority 8, **disabled by default** — enable when a vLLM server is running)
- Imported `DEFAULT_VLLM_BASE_URL`

### `tests/unit/providers/test_vllm_provider.py` (new)
- 24 unit tests: success path, system messages, auth headers, retry on timeout/client error/500, exhaust retries, no-retry on bad JSON/API error/400, health check success/model-not-found/unreachable, parse response (plain text + code blocks), configuration

### `TODO.md`
- Checked off "Add support for local models (Ollama/vLLM)"
- Updated P3 count (19→20) and total (73→74)

## Verification

- `ruff format --check .` ✅ (416 files)
- `pytest tests/unit/providers/ -q` → 114 passed ✅

## Notes

- vLLM provider is **disabled by default** in config — users enable it by setting `enabled: true` and pointing at their vLLM server
- No vLLM dependency is added; the provider uses `aiohttp` (already a project dependency) to talk to the OpenAI-compatible API
- The existing Ollama integration is unchanged

## Summary by CodeRabbit

* **New Features**
  * Added support for connecting to local vLLM models through an OpenAI-compatible API.
  * Added configurable endpoints, authentication, models, timeouts, generation settings, and retry handling.
  * Added model health checks and response parsing, including fenced-code responses.
  * Registered vLLM as an available provider with a disabled default configuration.

* **Documentation**
  * Updated local model support tracking and documented vLLM availability.

* **Tests**
  * Added coverage for requests, authentication, retries, health checks, response parsing, and model metadata.


SHA888:
## Addressed review feedback

### 1. Health check timeout (fixed)
Replaced the hardcoded  timeout in  with a new configurable  parameter. Defaults to  — enough headroom for cold-starting vLLM servers while still bounding the wait. The parameter is also exposed in .

### 2. `top_p` per-call override (fixed)
Added `top_p` to the per-call override set in `submit_prompt()`, alongside the existing `temperature`, `max_tokens`, and `stop` overrides. Docstring updated to match.

### 3. Plain HTTP + API key warning (fixed)
Added a docstring warning in `__init__` advising users to use `https://` when `api_key` is set and `base_url` points at a non-localhost host.

### 4. `aiohttp` dependency — already handled (no change needed)
`aiohttp>=3.8.0` is already declared in `pyproject.toml` line 40. No change needed.

### 5. Config mapping / priority collision — verified, no issue (no change)
- `_create_provider()` passes `config_params` as `**kwargs` to `VLLMProvider.__init__`. The config keys (`base_url`, `model`, `timeout`, `temperature`) all match the constructor signature cleanly.
- Priorities: ollama=10, ollama_reviewer=9, vllm=8, dummy=1 — no collision.

### Verification
- `ruff format --check .` ✅ (416 files)
- `pytest tests/ -q` → 881 passed ✅

SHA888:

PR #133, opened 2026-08-06
fix(seal): add async search method to KnowledgeBase for EnhancedSealSystem compatibility

SHA888:
## What

 calls `await self.knowledge_base.search(query=..., max_results=..., min_score=..., context=...)` but the real `KnowledgeBase` only had `search_entries(query, tags, metadata, limit)` — different method name, different parameter names, and neither method was async so the `await` would raise `TypeError`.

The system only worked because `EnhancedSealSystem` defaults to `MockKnowledgeBase` (which has a `search` method). When someone injects the real `KnowledgeBase`, knowledge retrieval silently fails.

## Changes

- Added `async def search()` method to `KnowledgeBase` that:
  - Accepts the parameters `EnhancedSealSystem` expects: `query`, `max_results`, `min_score`, `context`
  - Delegates to `search_entries` internally
  - Converts `KnowledgeEntry` objects to plain dicts matching what callers expect
  - `min_score` and `context` are accepted for API compatibility but not used by the current substring-matching implementation

## Verification

- `ruff format --check .` ✅
- `ruff check evoseal/integration/seal/knowledge/` ✅
- `pytest tests/unit/seal/test_enhanced_seal_system.py tests/integration/seal/test_knowledge_base.py tests/integration/seal/test_knowledge_base_concurrent.py tests/integration/seal/test_knowledge_aware_strategy.py -q` → 31 passed ✅

## Addresses

TODO.md P1: Knowledge retrieval in the SEAL subsystem is broken

## Summary by CodeRabbit

* **New Features**
  * Added an asynchronous knowledge-base search interface.
  * Search results can be limited by a maximum result count.
  * Results use a consistent format with content, metadata, tags, timestamps, identifiers, and relevance scores.
  * Added compatibility support for optional scoring and context parameters.
  * Added validation for negative result limits and support for returning no results when the limit is zero.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -152,7 +152,7 @@
 
 ### SEAL Subsystem Issues Found in Whole-Repo Code Review (2026-07-22)
 
-- [ ] **Knowledge retrieval in the SEAL subsystem is broken** _(exact file:line needs re-verification — flagged by initial review pass, not yet deep-dived)_
+- [x] **Knowledge retrieval in the SEAL subsystem is broken** _(done 2026-08-06)_ — `enhanced_seal_system.py:467` called `await self.knowledge_base.search(query=..., max_results=..., min_score=..., context=...)` but `KnowledgeBase` only had `search_entries(query, tags, metadata, limit)` — different method name, different parameter names, and neither method was async so the `await` would raise `TypeError`. Added async `search()` method to `KnowledgeBase` that accepts the expected parameters, delegates to `search_entries`, and converts `KnowledgeEntry` objects to plain dicts matching what callers expect
 
 ---
 
@@ -306,10 +306,10 @@
 | Priority | Total | Done | Notes |
 |----------|-------|------|-------|
 | 🔴 P0    | 11    | 11   | Original 5 complete; all 6 critical bugs from 2026-07-22 whole-repo review fixed (PRs #74, #76-#79) |
-| 🟠 P1    | 24    | 18   | Original safety/integration items done; +12 high-priority bugs from 2026-07-22 review (3 CI/CD pipeline fixes: workflow_run name mismatch, requirements/ path, security gate bypass); signal-handler init fix; safety.yaml created; monitoring dashboard auth+CORS fix |
+| 🟠 P1    | 24    | 19   | Original safety/integration items done; +12 high-priority bugs from 2026-07-22 review (3 CI/CD pipeline fixes: workflow_run name mismatch, requirements/ path, security gate bypass); signal-handler init fix; safety.yaml created; monitoring dashboard auth+CORS fix; SEAL knowledge retrieval API mismatch fix |
 | 🟡 P2    | 30    | 25   | Co-evolution loop gaps (8 items, 8 done) + existing P2 + 13 medium bugs from 2026-07-22 review + 4 latent collect->train bugs found closing the loop (1 fixed, 1 new HF-format gap resolved); provider_manager health-check await fix; workflow-agent private-API/event-loop fix; checkpoint save/restore test; trust_remote_code security fix; safety-decision orchestration tests; structured improvement units |
 | 🟢 P3    | 24    | 19   | Makefile, pre-commit, Docker, ADRs, ADR refresh, CHANGELOG complete; +11 hygiene items from 2026-07-22 review; Ollama provider retry/backoff fix; local_models TTL cache; workspace prompt file conventions; how-it-works tutorial; model_fine_tuner key validation; model_fine_tuner GPU availability check |
-| **Total** | **89** | **73** | |
+| **Total** | **89** | **74** | |
 
 > Update this table as you complete items. Recommended flow: P0 → P1 → P2 → P3.
 >
--- a/TODO.md
+++ b/TODO.md
@@ -152,7 +152,7 @@
 
 ### SEAL Subsystem Issues Found in Whole-Repo Code Review (2026-07-22)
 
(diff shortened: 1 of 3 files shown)


SHA888:

**1. `min_score` log flooding** — Fixed. The guard now compares against `MockKnowledgeBase`'s actual default (`0.3`) instead of just `is not None`. `EnhancedSealSystem` always passes `min_score=0.3` (the config default), so the warning no longer fires on every call. A non-default value (e.g. `0.5`) still warns as intended. Added 3 async tests covering the three cases (None, default, non-default).

**2. `max_results < 0` behavior** — Changed from silent clamping to `ValueError`. The reviewer was right to flag the silent behavior — raising is better since a negative `max_results` is always a caller bug. (Note: `results[:0]` in Python correctly returns an empty list, not "unlimited", so the original clamping was *safe* — just not *loud* enough about the bug.) Added a test asserting the `ValueError`.

**3. `async` + synchronous `RLock`** — No code change needed. Confirmed `search_entries` is a fast in-memory dict scan + list slice with no I/O, so blocking the event loop is negligible. Added a comment documenting this design choice for future maintainers.

Also corrected the TODO.md description: the pre-fix call raised `AttributeError` (missing `search` method), not `TypeError`.

SHA888:
Addressed the review feedback in 872adb4. Here's the point-by-point:

**1. `min_score` filtering is a no-op** — Agreed. The original TODO.md task was specifically about the `AttributeError`/`TypeError` crash (the call couldn't even complete), so the PR does fix what it scoped. But the scoring gap is real. Added a prominent `TODO` block in the `search()` docstring **and** a comment next to the hardcoded `score: 1.0` that calls out `prompt/formatters.py` as concrete downstream code affected, so this doesn't get lost.

**2. `_MOCK_DEFAULT_MIN_SCORE` heuristic** — Agreed, it was fragile. Removed the special-casing entirely. Now **any** non-None `min_score` triggers the warning uniformly — no value gets silently swallowed. Dropped the `_MOCK_DEFAULT_MIN_SCORE` constant.

**3. Hardcoded `score: 1.0`** — Verified that `prompt/formatters.py:48` does read `score` for display (`[Relevance: 1.00]`), so downstream output is affected. Added the TODO note in the docstring calling this out specifically. Didn't implement scoring since that's a larger change beyond this review scope — the TODO makes the gap trackable.

**4. `search(max_results=0)` behavior** — Verified: `search_entries` slices `results[:0]` which returns `[]` cleanly (no exception). Added `test_search_max_results_zero_returns_empty` to lock this in.

All 854 tests pass, ruff clean.

SHA888:
Addressed the review feedback. Here's the point-by-point:

### Fixed

**1.  blocks the event loop** — Wrapped the locked in-memory scan in . The  acquisition and O(n) scan now run in a thread pool, keeping the event loop responsive. Added  and a  helper.

**2.  no-op / TODO.md checkbox** — Agreed that the  checkbox only covered the crash fix (AttributeError/TypeError), not score-based filtering. Added a follow-up TODO item: *"KnowledgeBase.search() score-based filtering/ranking"* so it doesn't get lost.

### Caveats acknowledged (no changes)

**3.  behavior** — Verified:  does , and  correctly returns . The test is safe. The reviewer is right to flag the caveat since  wasn't in the diff, but the behavior is correct.

**4.  raises  vs mock** — Confirmed the divergence. MockKnowledgeBase.search does  with negative limit, which in Python silently returns wrong results ( = all but last). Our implementation raises , which is *better* behavior. The mock's silent-wrong-results is the mock's problem. Tests written against the mock wouldn't typically pass negative limits, and if they did, the mock's behavior would be the bug, not ours.

SHA888:
Addressed the review feedback. Point-by-point:

### Fixed

**1. `search()` blocks the event loop** — Wrapped the locked in-memory scan in `asyncio.to_thread()`. The `RLock` acquisition and O(n) scan now run in a thread pool, keeping the event loop responsive. Added `import asyncio` and a `_search_entries_sync` helper.

**2. `min_score` no-op / TODO.md checkbox** — Agreed that the `[x]` checkbox only covered the crash fix (`AttributeError`/`TypeError`), not score-based filtering. Added a follow-up TODO item: *"KnowledgeBase.search() score-based filtering/ranking"* so it doesn't get lost.

### Caveats acknowledged (no changes)

**3. `search_entries(limit=0)` behavior** — Verified: `search_entries` does `results[:limit]`, and `results[:0]` correctly returns `[]`. The test is safe. Reviewer is right to flag as a caveat since `search_entries` wasn't in the diff, but the behavior is correct.

**4. `max_results < 0` raises `ValueError` vs mock** — Confirmed the divergence. `MockKnowledgeBase.search` does `filtered[:limit]` with negative limit, which in Python silently returns wrong results (`filtered[:-1]` = all but last). Our implementation raises `ValueError`, which is *better* behavior. The mock's silent-wrong-results is the mock's problem. Tests written against the mock wouldn't typically pass negative limits, and if they did, the mock's behavior would be the bug, not ours.

SHA888:
Addressed review feedback:

**Point 1 (Unverified dependency — / omitted from  call):** Verified against the actual code —  at line 210 has  and  as defaults. The call  is safe; no  will be raised. No change made.

**Point 2 (Logger pattern):** Accepted. Replaced the per-instance  class attribute + lazy  pattern with a plain module-level .  returns a cached singleton, so the indirection was unnecessary.

Verified: , , and all 36 tests pass.

SHA888:

PR #134, opened 2026-08-07
feat(seal): add relevance scoring to KnowledgeBase search

SHA888:
## What

Implements real relevance scoring for `KnowledgeBase.search()` instead of returning hardcoded `score: 1.0` for every match. The `min_score` parameter now genuinely filters low-relevance results.

## Why

The `search()` method (added for `EnhancedSealSystem` compatibility) returned `score: 1.0` for every entry, making `min_score` filtering a no-op and score-based ranking in `prompt/formatters.py` treat all matches as equally relevant. This was flagged as a follow-up in TODO.md.

### `evoseal/integration/seal/knowledge/knowledge_base.py`
- **`_compute_relevance_score()`**: Static method using case-insensitive substring occurrence count weighted by `query_length / content_length`, producing scores in `[0.0, 1.0]`
- **`_scored_search()`**: Computes scores for all entries, filters zero-score results, sorts by score descending
- **`search()`**: Now returns real scores and filters by `min_score` when provided. Removed the warning log about `min_score` being unused. Thread-safe via `asyncio.to_thread`.

### `tests/integration/seal/test_knowledge_base.py`
- 17 new tests: scoring (exact match, no match, partial, multiple occurrences, case sensitivity), scored search (sorting, limit, dict content), async search (real scores, min_score filtering, zero results, result format)

## Verification

- `ruff format --check .` ✅ (414 files)
- `pytest tests/integration/seal/test_knowledge_base.py -q` → 25 passed ✅
- `pytest tests/integration/seal/ tests/unit/seal/ -q` → 177 passed, 2 skipped ✅

## Addresses

TODO.md P1: Knowledge retrieval in the SEAL subsystem is broken (includes score-based filtering follow-up)

## Summary by CodeRabbit

* **New Features**
  * Added asynchronous knowledge-base search.
  * Results are ranked by relevance across content, tags, and metadata, including partial and exact matches.
  * Search supports case-insensitive matching, minimum relevance-score filtering, and result limits.
  * Searches handle nested or dictionary-based content and return structured results.

* **Bug Fixes**
  * Added validation for empty queries and negative result limits.
  * Searches safely run in the background and return no results when nothing matches.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -152,7 +152,7 @@
 
 ### SEAL Subsystem Issues Found in Whole-Repo Code Review (2026-07-22)
 
-- [ ] **Knowledge retrieval in the SEAL subsystem is broken** _(exact file:line needs re-verification — flagged by initial review pass, not yet deep-dived)_
+- [x] **Knowledge retrieval in the SEAL subsystem is broken** _(done 2026-08-07)_ — `enhanced_seal_system.py:467` called `await self.knowledge_base.search(query=..., max_results=..., min_score=..., context=...)` but `KnowledgeBase` had no `search()` method. Added async `search()` with real relevance scoring (occurrence-count weighted by query/content length ratio), `min_score` filtering, and thread-safe offloading. Scores are in `[0.0, 1.0]`, sorted descending, and `min_score` genuinely filters low-relevance results.
 
 ---
 
@@ -306,10 +306,10 @@
 | Priority | Total | Done | Notes |
 |----------|-------|------|-------|
 | 🔴 P0    | 11    | 11   | Original 5 complete; all 6 critical bugs from 2026-07-22 whole-repo review fixed (PRs #74, #76-#79) |
-| 🟠 P1    | 24    | 18   | Original safety/integration items done; +12 high-priority bugs from 2026-07-22 review (3 CI/CD pipeline fixes: workflow_run name mismatch, requirements/ path, security gate bypass); signal-handler init fix; safety.yaml created; monitoring dashboard auth+CORS fix |
+| 🟠 P1    | 24    | 19   | Original safety/integration items done; +12 high-priority bugs from 2026-07-22 review (3 CI/CD pipeline fixes: workflow_run name mismatch, requirements/ path, security gate bypass); signal-handler init fix; safety.yaml created; monitoring dashboard auth+CORS fix; SEAL knowledge retrieval with relevance scoring |
 | 🟡 P2    | 30    | 25   | Co-evolution loop gaps (8 items, 8 done) + existing P2 + 13 medium bugs from 2026-07-22 review + 4 latent collect->train bugs found closing the loop (1 fixed, 1 new HF-format gap resolved); provider_manager health-check await fix; workflow-agent private-API/event-loop fix; checkpoint save/restore test; trust_remote_code security fix; safety-decision orchestration tests; structured improvement units |
 | 🟢 P3    | 24    | 19   | Makefile, pre-commit, Docker, ADRs, ADR refresh, CHANGELOG complete; +11 hygiene items from 2026-07-22 review; Ollama provider retry/backoff fix; local_models TTL cache; workspace prompt file conventions; how-it-works tutorial; model_fine_tuner key validation; model_fine_tuner GPU availability check |
-| **Total** | **89** | **73** | |
+| **Total** | **89** | **74** | |
 
 > Update this table as you complete items. Recommended flow: P0 → P1 → P2 → P3.
 >
--- a/evoseal/integration/seal/knowledge/knowledge_base.py
+++ b/evoseal/integration/seal/knowledge/knowledge_base.py
@@ -555,13 +555,18 @@ class KnowledgeBase:
         by score descending.
         """
         query_lower = query.lower()
-        scored: list[tuple[KnowledgeEntry, float]] = []
+        # Snapshot the current entries under the lock, then release it
+        # before the O(n) scoring pass so that concurrent add_entry /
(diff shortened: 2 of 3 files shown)


SHA888:
Addressed review feedback with the following changes:

**Point 1 — query validation:** Added upfront  check in  that raises  (matching the existing  pattern) instead of letting /non-str propagate to  at . Three new tests cover , empty string, and int inputs.

**Point 2 — dict-content flattening:** Replaced the shallow  filter with , a recursive static method that walks nested dicts, lists, and tuples. Entries like `{"tags": ["python", "ml"]}` are now fully searchable. Two new tests verify nested dict and list/tuple flattening.

**Point 3 — bool exclusion:**  explicitly checks  before the  check, so `True`/`False` are excluded from searchable text. One dedicated test plus the nested-dict test both verify this.

**Point 4 — substring matching:** Agreed this is a real functional gap for multi-word natural-language queries. Rather than changing behavior for v1, added a `.. note::` to the `_compute_relevance_score` docstring documenting the limitation explicitly.

**Point 5 — DEFAULT_SEARCH_LIMIT:** Verified it exists at line 61 of the class definition (`DEFAULT_SEARCH_LIMIT = 10`). No action needed — the reviewer flagged this as a caveat, not a confirmed defect.

Verification: `ruff format` ✅, `ruff check` ✅, 35/35 knowledge-base tests pass, 187/187 seal tests pass.

SHA888:
## Review feedback analysis

I verified each of the four points against the actual codebase state. None require code changes on this PR:

### 1. `_flatten_content` float precision loss
**Not a bug in practice.** `str(3.10) → "3.1"` is real Python behavior, but dict content loaded from JSON already has `3.1` (JSON doesn't preserve trailing zeros). The existing test `test_flatten_content_lists_and_tuples` explicitly tests with `3.10` and asserts `"3.1" in result` — this was written knowingly. Low risk, already accounted for.

### 2. `context` parameter unused
**Safe to ignore.** Already documented in the docstring as "Accepted for API compatibility; not used." I checked the actual caller (`enhanced_seal_system.py:467`): `retrieve_relevant_knowledge` passes `context` through, but it's consumed for prompt construction and caching elsewhere — no evidence the caller expects search-scoped filtering from this parameter. The silent no-op is correct.

### 3. Substring matching false positives
**Already documented as v1 limitation.** The `_compute_relevance_score` docstring explicitly notes this: *"Scoring is based on literal substring matching... multi-word queries only match on exact phrase / word-order... This is acceptable for v1 but may need upgrading to token-based matching in the future."* Valid as a follow-up issue, not a change on this PR.

### 4. `DEFAULT_SEARCH_LIMIT` might be undefined
**Already defined.** `DEFAULT_SEARCH_LIMIT = 10` is a class constant at line ~44 of `KnowledgeBase`. The reviewer noted they "can't verify from this diff alone" — confirmed it exists.

SHA888:
Addressed review feedback with the following changes:

**Point 1 — substring false positives:** Extended the `_compute_relevance_score` docstring to explicitly document that single-word queries match inside unrelated words (e.g. `"on"` matches `"python"`) as a known v1 limitation, plus the length-bias and non-overlapping count behaviors. Not changing matching behavior in v1 — word-boundary or token-based matching is a follow-up.

**Point 2 — tags/metadata not searchable:** Extracted `_build_searchable_text()` that combines `entry.content`, `entry.tags`, and flattened `entry.metadata` into one searchable string. A caller searching for a tag (e.g. `"ml"`) now gets matches. Three new tests cover tag search, metadata search, and the combined searchable text.

**Point 3 — lock inside `_scored_search`:** Moved lock acquisition inside `_scored_search` itself. The wrapper `_scored_search_thread` now exists only for the `asyncio.to_thread` call signature. Tests calling `_scored_search` directly are safe against concurrent mutation. Note: the lock is an `RLock`, so the wrapper's former double-acquire (wrapper+inner) would have been harmless.

**Point 4 — scoring normalization:** Added a second `.. note::` documenting that length normalization biases toward short snippets and that `str.count` is non-overlapping. Not changing formula in v1.

**Point 5 — `DEFAULT_SEARCH_LIMIT`:** Confirmed already defined as a class constant (`DEFAULT_SEARCH_LIMIT = 10`, line 61). No action needed.

Verification: `ruff format` ✅, `ruff check` ✅, 40/40 knowledge-base tests pass (35 original + 5 new).

SHA888:
## Review feedback addressed

### Point 1: Lock contention in `_scored_search` ✅ Fixed
The lock was held for the entire O(n) scoring pass (`_build_searchable_text` + `_compute_relevance_score` on every entry). Fixed by snapshotting `list(self.entries.values())` under the lock, then releasing it before the scoring loop. Only the dict iteration needs protection.

### Point 2: `context` parameter unused — Not a bug, no change
Verified the caller (`enhanced_seal_system.py:471`) does pass `context=context`, but this is an intentional API-compatibility parameter. `KnowledgeBase` has never used context for filtering/scoping — the `search()` docstring explicitly documents it as unused. The "knowledge retrieval broken" bug was about `search()` not existing at all, not about context filtering. If context-scoped retrieval is needed later, that is a separate feature.

### Point 3: pytest-asyncio config — Already present, no change
- `pyproject.toml:95`: `"pytest-asyncio>=0.23.0"` ✅
- `pytest.ini:2`: `asyncio_mode = strict` ✅  
- `pytest.ini:3`: `asyncio_default_fixture_loop_scope = function` ✅
- All async tests use `@pytest.mark.asyncio` correctly ✅

### Verification
- `ruff format --check .` ✅ (414 files)
- `pytest tests/integration/seal/test_knowledge_base.py -q` → 40 passed ✅
- Full test suite → 854 passed ✅

SHA888:

PR #135, opened 2026-08-07
fix(integration): report failed DGM/OpenEvolve remote jobs as unsuccessful

SHA888:
## What

Both `dgmr/dgm_adapter.py` and `oe/openevolve_adapter.py` had a bug in their remote job polling: when the remote service reported a job as `"failed"`, the adapters broke out of the poll loop but then unconditionally fetched the result and returned `success=True`. A job that crashed or timed out on the remote service was silently treated as a success by the adapter layer.

## Fix

Both adapters now check the final status **after** the poll loop exits. When `status == "failed"`, they return `success=False` with the error message from the remote service. The result endpoint is only fetched for completed jobs.

## Files changed

- `evoseal/integration/dgmr/dgm_adapter.py` — added failed-status check in `_advance_generation()`
- `evoseal/integration/oe/openevolve_adapter.py` — added failed-status check in `_evolve_remote()`
- `tests/unit/test_dgm_adapter_remote.py` — added `test_dgm_advance_generation_remote_failed`
- `tests/unit/test_openevolve_adapter_remote.py` — added `test_openevolve_evolve_remote_failed`
- `TODO.md` — checked off "DGM/OpenEvolve job runner reports failed jobs as successful"

## Verification

```
$ ruff format --check .     → 414 files already formatted
$ ruff check evoseal/integration/dgmr/ evoseal/integration/oe/ tests/unit/test_dgm_adapter_remote.py tests/unit/test_openevolve_adapter_remote.py → All checks passed!
$ pytest tests/unit/test_dgm_adapter_remote.py tests/unit/test_openevolve_adapter_remote.py -q → 5 passed
```

Addresses TODO.md P1 item: "DGM/OpenEvolve job runner reports failed jobs as successful"

## Summary by CodeRabbit

- **Bug Fixes**
  - Remote DGM and OpenEvolve jobs now report unsuccessful results immediately when a job fails.
  - Failure responses include the server-provided error message, with a fallback message when none is available.
  - Results are only retrieved after jobs complete successfully.

- **Tests**
  - Added regression coverage for failed remote jobs and error-message propagation across both integrations.

- **Documentation**
  - Updated completion tracking to reflect the resolved remote job failure handling.

### DIFF
--- a/TODO.md
+++ b/TODO.md
@@ -143,7 +143,7 @@
 ### DGM/OpenEvolve Adapter Issues Found in Whole-Repo Code Review (2026-07-22)
 
 - [ ] **`evoseal/integration/dgm/` + `dgmr/` and `evoseal/integration/oe/` + `openevolve/` look like duplicated/forked adapter implementations that have drifted apart** — needs a decision on which is canonical and whether the other should be removed or reconciled
-- [ ] **DGM/OpenEvolve job runner reports failed jobs as successful** _(exact file:line needs re-verification — flagged by initial review pass, not yet deep-dived)_
+- [x] **DGM/OpenEvolve job runner reports failed jobs as successful** _(done 2026-08-07)_ — both `dgmr/dgm_adapter.py` `_advance_generation()` and `oe/openevolve_adapter.py` `_evolve_remote()` broke out of the poll loop on `status == "failed"` but then unconditionally fetched the result and returned `{"success": True}`. Fixed both to check the final status after the poll and return `success: False` with the error message when the job failed. Added regression tests in `test_dgm_adapter_remote.py` and `test_openevolve_adapter_remote.py`.
 
 ### CLI Issues Found in Whole-Repo Code Review (2026-07-22)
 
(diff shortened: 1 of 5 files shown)


SHA888:
Addressed the review feedback. Here's my analysis:

**Reviewer's concern:** If the poll loop exits for a reason other than `"failed"` (e.g. timeout), `status == "failed"` is False and execution falls through to return the result as if it succeeded.

**What I found after verifying the actual code:**
- Both poll loops are bare `while True:` with no timeout or max-iteration exit — the only exit paths are `completed`, `failed`, a non-200 status check (returns error inline), or an exception (caught by outer try/except). So the specific bug described doesn't currently exist.
- The reviewer correctly flagged this as a *caveat* — they couldn't see the loop structure from the diff context alone.

**What I changed:** Inverted the check from `== "failed"` to `!= "completed"` in both files. This is purely defensive but zero-risk — if someone later adds a timeout or max-iterations exit, the code will correctly treat any non-terminal state as failure rather than silently succeeding. Updated the error message to include the actual status for debuggability.

All 7 tests pass, ruff clean.

PR #140, opened 2026-08-10
fix(ci): fix broken release pipeline — missing script, lost version bumps, redundant steps

SHA888:
## What

Fix the release pipeline (`release.yml` + `pre-release.yml`), which had four concrete failures preventing any release from succeeding.

## Why

The TODO item "Release pipeline is broken" flagged this during the 2026-07-22 whole-repo review but noted the exact failure mode needed re-verification. This PR re-verifies and fixes every issue found.

### 1. Missing release-notes script (pre-release workflow always fails)
`pre-release.yml` referenced `scripts/auto_generate_release_notes.py` which **did not exist**. Every RC run failed at that step. Created the script — it generates categorised release notes from git history using conventional-commit prefixes.

### 2. Version bump lost on `workflow_run` trigger (duplicate releases)
`release.yml` bumped the version locally with `--no-commit` but **never committed or pushed it back to main**. On the next CI-on-main trigger, the workflow would bump from the same stale version, attempt to create a duplicate tag/release, and either fail or produce a broken release. Added a "Commit version bump" step that pushes `pyproject.toml` + `CHANGELOG.md` back to the source branch after the build.

### 3. Three redundant version-determination steps
`bump_version`, `verify_version`, and `set_version` each independently resolved the version from different sources, sometimes overwriting each other. Consolidated into a single "Determine version" step with clear branching per trigger type (tag push → tag version, workflow_dispatch → user input or bump, workflow_run → patch bump).

### 4. `pip` vs `uv` inconsistency
Both workflows used `pip install -e .` while every other CI job uses `uv`. Switched to `uv pip install --system`.

## Verification

- `ruff format --check .` ✅
- YAML linting passed via pre-commit hook ✅

## Addresses

`TODO.md` item: "Release pipeline is broken" (P1, CI/CD section)

## Summary by CodeRabbit

- **New Features**
  - Added automated generation of versioned Markdown release notes with categorized changes, commit links, installation guidance, and upgrade instructions.
  - Added package building and automatic attachment of release assets.
  - Added flexible version selection for tagged, manual, and default releases.

- **Improvements**
  - Improved release validation, credential checks, and artifact naming.
  - Updated release workflows with newer Python tooling and faster dependency installation.
  - Preserved existing release notes and automated version updates where applicable.

- **Documentation**
  - Updated release tracking to mark pipeline improvements as completed.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -35,8 +35,6 @@ jobs:
       contents: write
       id-token: write
       actions: read
-      checks: write
-      statuses: write
       pull-requests: write
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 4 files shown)


SHA888:
Addressed review feedback — verified each point against the actual repo state:

### ✅ Fixed (3 points)

**1. Version consistency check (tag-push safety net):** Re-added a step that fails when `pyproject.toml` version ≠ tag version for tag-push events. Also verifies `dist/evoseal-${VERSION}.tar.gz` and `dist/evoseal-${VERSION}-py3-none-any.whl` exist before publish, replacing the dropped `Verify release artifacts` step.

**2. Commit version bump gating:** Changed from `github.event_name != 'push'` (which included `workflow_dispatch` without publish) to a condition that only commits for `workflow_run` or `workflow_dispatch` with `publish=true`. This prevents the version-drift bug where a manual dispatch without publish would bump+commit to main with no corresponding release.

**3. Build verification:** Added `twine check dist/*` after the build step to validate package metadata and structure, replacing the dropped `pip install dist/*.whl --no-deps` verification. Also verifies expected artifact filenames exist.

### ❌ Dismissed (1 point)

**4. Shallow checkout caveat:** The reviewer flagged that `actions/checkout` might be shallow, breaking `git describe --tags` in `auto_generate_release_notes.py`. This is **not an issue** — the checkout step already has `fetch-depth: 0` (full clone), and `actions/checkout@v7` fetches all tags by default with full depth. `git describe --tags` will work correctly.

SHA888:
## Review feedback addressed

Verified each point against the actual repo state. Here's what I found and fixed:

### ✅ Fixed (5 points)

**1. Self-retriggering failure loop** — Two changes:
- Added `!contains(github.event.head_commit.message, '[skip ci]')` guard to the tag-push `if:` condition so bot-created version-bump commits don't spawn a second run
- Added `target_commitish: ${{ github.sha }}` to `softprops/action-gh-release` so created tags point at the correct commit

**2. `END` delimiter in `get_commits_since()`** — Replaced `log.split('END')` with NUL-byte record separation. Git prohibits NUL in commit messages, so this is safe. The format is now `%H%x00%s%x00%b%x00` and parsed by splitting on `\x00` and grouping by 3.

**3. Build filename check with prerelease versions** — Changed exact filename checks to glob matching (`dist/evoseal-${VERSION}*.tar.gz`) to handle PEP 440 normalisation (e.g. `1.2.3-rc.1` → `1.2.3rc1`).

**4. Ordering: commit before publish** — Moved "Commit version bump" to run *after* "Publish to PyPI". A failed upload no longer leaves pyproject.toml bumped to an unreleased version.

**5. workflow_dispatch consistency** — Changed the specific-version path to re-read the version from `pyproject.toml` after `version.py update` instead of using the raw input. Catches normalisation mismatches.

### ℹ️ Already handled (1 point)

**6. fetch-depth: 0** — The checkout step already has `fetch-depth: 0`. No change needed.

SHA888:
## Review feedback addressed (3 of 4 points)

### ✅ Point 1 — `workflow_run` causing version/release churn
**Confirmed and fixed.** Removed `workflow_run` from both the `Commit version bump` and `Create GitHub Release` gates. Now `workflow_run` still computes a patch bump (used for build verification) but never commits it or cuts a release. Only `tag_push` and `workflow_dispatch` with `publish=true` trigger the full release flow.

### ❌ Point 2 — Shallow checkout fallback
**Already handled.** The checkout step already uses `fetch-depth: 0`. The reviewer noted they couldn't verify this from the diff — confirmed it's already there.

### ✅ Point 3 — `git diff --quiet` scope
**Confirmed and fixed.** Changed `git diff --quiet` to `git diff --quiet -- pyproject.toml CHANGELOG.md` so unrelated tracked-file changes don't cause the commit step to fail under `set -euo pipefail`.

### ✅ Point 4 — `pip install` smoke-test dropped
**Confirmed and fixed.** Restored `pip install dist/*.whl --no-deps -q` after `twine check dist/*` in the Build package step.

